PR #2, opened 2023-04-15
mimc7 example implementation

qwang98:
Adapted from circomlib mimc7 implementation here: https://github.com/iden3/circomlib/blob/master/circuits/mimc.circom


qwang98:
Replace by pul request #11

PR #4, opened 2023-04-21
MiMC7 circuit example

qwang98:
updates since last review:
1. added two lookup columns (row index, round constant) using the latest lookup syntax
2. enforced lookup columns match to source columns
3. enforced row index to 0 for the first step and to increment by one for all other steps
4. added new step type for the first step, in order to constrain row index to 0
5. added trace inputs


qwang98:
Note that I committed the MiMC7 sponge example by mistake. Let me know if there's a way to skip this file for the merge.

leolara:
@qwang98 @10to4 this PR has changess to files that it should not, like ast.rs, please clean that and let me know so I review it.

qwang98:
Will do! Those are from the lookup column that weren't merged yet as of the PR creation.

> @qwang98 @10to4 this PR has changess to files that it should not, like ast.rs, please clean that and let me know so I review it.

qwang98:
Replaced by pull request #11 . Closing this now.

PR #13, opened 2023-05-01
Steve/issue 6 api documentation v2

qwang98:
see commit message :)

API documentation deployed on fork here: https://qwang98.github.io/chiquito/chiquito/index.html


ChihChengLiang (changes requested):
We should add the `docs/` directory to the `.gitignore` file so that we don't check in the directory to the codebase. 
We don't check in the `docs` since the contents are autogenerated. No need to review the generated content now or in the future so we can ignore them.

qwang98:
> 

Hi CC! (let me know if you prefer another name)

Thanks for the feedback. I'd love to point out that docs subfolder (or a customized folder path that's pushed to a branch) is required by Github pages to deploy the docs, so I can think of two solutions:
1. If we don't want docs to appear in the main, we can deploy it to another branch, but this would require maintaining two branches, and we still can't gitignore the folder on the branch
2. I can deploy it on my fork, but the URL will have my github user name instead of pse on it
3. I'm not sure if this works but there might be a CI job that skips checks for some paths?

Let me know which way you think is the best!

ChihChengLiang:
Hi @qwang98,

GitHub page can render static contents in a folder on the `gh-pages` branch. We can follow the approach of this configuration https://github.com/zcash/zcash/blob/master/.github/workflows/book.yml.
Note that the script builds static files using mdbook. We can replace mdbook with cargo doc to build our static files.

qwang98:
> Hi @qwang98,
> 
> GitHub page can render static contents in a folder on the `gh-pages` branch. We can follow the approach of this configuration https://github.com/zcash/zcash/blob/master/.github/workflows/book.yml. Note that the script builds static files using mdbook. We can replace mdbook with cargo doc to build our static files.
> 
> (CC is a perfect name, feel free to use it)

Thanks CC. I created another workflow for auto deployment of docs static files to the gh-pages branch. Docs should now be auto updated whenever there's a push to the main.

@leolara Can you set up the gh page on PSE chiquito repository using the screenshot I sent you over TG? I don't have permission. Thanks!

src/dsl/cb.rs
```diff
@@ -209,27 +269,49 @@ pub fn if_next_step<F: Clone, T: Into<Constraint<F>>>(
...
+/// # **Return:**
+/// A **`Constraint<F>`** representing the requirement that the next step must not be of the given step type.
 pub fn next_step_must_not_be<F: From<u64>>(step_type: StepTypeHandler) -> Constraint<F> {
     annotate(
         format!("next_step_must_be({})", step_type.annotation),
         step_type.next(),
     )
 }
 
+/// The **`annotate`** function takes a string annotation and an expression, and returns a new constraint with the given annotation and expression.
```
ChihChengLiang, line 296: We can omit the subject in the sentence because the document will render them. Removing the subject can make the document look cleaner
Before:
Screenshot 2023-05-03 at 2 47 14 PM
After:
Screenshot 2023-05-03 at 2 49 01 PM


```suggestion
/// Take a string annotation and an expression, and returns a new constraint with the given annotation and expression.
```

src/dsl.rs
```diff
@@ -10,27 +10,65 @@ use core::fmt::Debug;
...
         Queriable::Halo2AdviceQuery(self.sc.add_halo2_advice(name, column), 0)
     }
 
+    /// # **Description:**
+    /// Imports a halo2 fixed column into the circuit and returns a **`Queriable`** instance.
+    /// # **Arguments:**
+    /// - **`name: &str`** - The name of the halo2 fixed column.
+    /// - **`column: Halo2Column<Fixed>`** - The fixed column to import.
+    /// # **Return:**
+    /// - **`Queriable<F>`** - A **`Queriable`** instance representing the imported halo2 fixed column.
```
ChihChengLiang, line 61: We should leave the main sentence in the first line. The document treats it specially. For example, when you search the function, the doc shows the first line. And if we put "Description" in the first line, the doc shows something unhelpful

Screenshot 2023-05-03 at 2 53 59 PM

We don't need to doc the params and the return types because we can jump and search them in the document.
```suggestion
    /// Imports a halo2 fixed column into the circuit and returns a **`Queriable`** instance.
```

ChihChengLiang (approved):
Nice, now the document is already reviewable.
I left some general editing guidelines, but feel free to ignore them in this PR.
The PR is already LGTM.

qwang98:
Thanks CC and Leo! I will update these plus conforming to the formatter changes in another PR.

src/dsl/cb.rs
qwang98: Fixed in pr#18

src/dsl.rs
qwang98: fixed in pr#18

PR #18, opened 2023-05-03
steve/issue_17-fix_ci_and_doc_format

qwang98:


qwang98:
> @qwang98 Do you use a case-insensitive file system like macOS's? Try ` git mv examples/MiMC7.rs examples/mimc7.rs`
> 
> See https://stackoverflow.com/a/24979063

Thanks CC! This worked like a charm and I should have asked earlier to save 2 hours trying to debug. Would just ask you to approve so that I can merge.

ChihChengLiang (approved):
LGTM. Great work addressing all the issues.

PR #19, opened 2023-05-04
finished all tests for constraint builder functions; not for lookups yet

qwang98:
Identified a few potential issues:
1. Input for `not` should be 0 or 1, but there's no enforcement of that
2. `xor` takes Into Expr and returns Into Expr, while almost all other functions take Into Constraint and return Into Constraint

Next steps:
1. I think annotations are broadly testable, it's just that the numbers are currently in hex format with padded zeroes, which are quite hard to read in the annotation messages as well. I think I should be able to fix it if you'd like to.

### DIFF
--- a/src/dsl/cb.rs
+++ b/src/dsl/cb.rs
@@ -244,7 +244,7 @@ pub fn next_step_must_be<F: From<u64>>(step_type: StepTypeHandler) -> Constraint
 /// the given step type.
 pub fn next_step_must_not_be<F: From<u64>>(step_type: StepTypeHandler) -> Constraint<F> {
     annotate(
-        format!("next_step_must_be({})", step_type.annotation),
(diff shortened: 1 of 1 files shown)


src/dsl/cb.rs
```diff
@@ -244,7 +244,7 @@ pub fn next_step_must_be<F: From<u64>>(step_type: StepTypeHandler) -> Constraint
 /// the given step type.
 pub fn next_step_must_not_be<F: From<u64>>(step_type: StepTypeHandler) -> Constraint<F> {
     annotate(
-        format!("next_step_must_be({})", step_type.annotation),
+        format!("next_step_must_not_be({})", step_type.annotation),
```
leolara, line 247: thanks for fixing this

leolara (approved):

PR #26, opened 2023-05-08
Removed leading zeroes for the debug msg of Expr and PolyExpr

qwang98:
- Only done for Const enum types, because other types don't involve leading zeroes (they can be composed of Const type though)
- Wrote test cases as well

I think the debug msg for the entire circuit is still from Halo2 native, and it doesn't remove leading zeroes. There's probably no way we can fix that unless we change the dependencies ourselves.


leolara:
@qwang98 I think we are getting that clippy error because we have a different config in .github that we have in local. I cannot reproduce those errors in local

PR #27, opened 2023-05-09
Add examples: poseidon hash

10to4:
Implementation Reference:
  https://github.com/HarryR/ethsnarks/blob/master/src/gadgets/poseidon.hpp

  https://github.com/iden3/circomlib/blob/master/circuits/poseidon_old.circom
  https://github.com/iden3/circomlib/blob/master/circuits/poseidon_constants_old.circom


leolara:
@10to4 have you tested with other poseidon hash implementation that it gives the same results for the same values?

leolara:
@10to4 can you put the constants at the bottom?

10to4:
> @10to4 have you tested with other poseidon hash implementation that it gives the same results for the same values?

A bit different. My implementation is based on the the gadgets in [ethsnarks](https://github.com/HarryR/ethsnarks/blob/master/src/gadgets/poseidon.hpp). However, I discovered that the parameters used in ethsnarks are generated randomly using the blake2b function. Implementing this in the context of chiquito would be more complex. We would need to first create the blake2b circuit, otherwise, we wouldn't be able to verify if the parameters are correct.

Later on, I went through code from other implementation versions, including circomlib (which has two versions: [old](https://github.com/iden3/circomlib/blob/master/circuits/poseidon_old.circom), [new](https://github.com/iden3/circomlib/blob/master/circuits/poseidon.circom)) and [go-iden3-crypto](https://github.com/iden3/go-iden3-crypto/blob/master/poseidon/poseidon.go), and noticed slight variations in each version of the Poseidon hash function. These versions differ somewhat from the one in ethsnarks. They all use fixed parameters, and the parameters in circom_old_poseidon and go-iden3-crypto are consistent. Consequently, I used this set of parameters in my example as well.

If there's a need to adjust the algorithm version, I can make modifications at any time.

10to4:
> @10to4 can you put the constants at the bottom?

committed.

leolara:
> > @10to4 have you tested with other poseidon hash implementation that it gives the same results for the same values?
> 
> A bit different. My implementation is based on the the gadgets in [ethsnarks](https://github.com/HarryR/ethsnarks/blob/master/src/gadgets/poseidon.hpp). However, I discovered that the parameters used in ethsnarks are generated randomly using the blake2b function. Implementing this in the context of chiquito would be more complex. We would need to first create the blake2b circuit, otherwise, we wouldn't be able to verify if the parameters are correct.
> 
> Later on, I went through code from other implementation versions, including circomlib (which has two versions: [old](https://github.com/iden3/circomlib/blob/master/circuits/poseidon_old.circom), [new](https://github.com/iden3/circomlib/blob/master/circuits/poseidon.circom)) and [go-iden3-crypto](https://github.com/iden3/go-iden3-crypto/blob/master/poseidon/poseidon.go), and noticed slight variations in each version of the Poseidon hash function. These versions differ somewhat from the one in ethsnarks. They all use fixed parameters, and the parameters in circom_old_poseidon and go-iden3-crypto are consistent. Consequently, I used this set of parameters in my example as well.
> 
> If there's a need to adjust the algorithm version, I can make modifications at any time.

I think if we are equivalent with circomlib it could be enough, but have you check we get the same result as in circomlib?

10to4:
> > > @10to4 have you tested with other poseidon hash implementation that it gives the same results for the same values?
> > 
> > 
> > A bit different. My implementation is based on the the gadgets in [ethsnarks](https://github.com/HarryR/ethsnarks/blob/master/src/gadgets/poseidon.hpp). However, I discovered that the parameters used in ethsnarks are generated randomly using the blake2b function. Implementing this in the context of chiquito would be more complex. We would need to first create the blake2b circuit, otherwise, we wouldn't be able to verify if the parameters are correct.
> > Later on, I went through code from other implementation versions, including circomlib (which has two versions: [old](https://github.com/iden3/circomlib/blob/master/circuits/poseidon_old.circom), [new](https://github.com/iden3/circomlib/blob/master/circuits/poseidon.circom)) and [go-iden3-crypto](https://github.com/iden3/go-iden3-crypto/blob/master/poseidon/poseidon.go), and noticed slight variations in each version of the Poseidon hash function. These versions differ somewhat from the one in ethsnarks. They all use fixed parameters, and the parameters in circom_old_poseidon and go-iden3-crypto are consistent. Consequently, I used this set of parameters in my example as well.
> > If there's a need to adjust the algorithm version, I can make modifications at any time.
> 
> I think if we are equivalent with circomlib it could be enough, but have you check we get the same result as in circomlib?

I have submitted the modifications, which gets currently the same result with circomlib. Before it was slightly different, because it was written according to ethsnarks' algorithm with circomlib’s parameters.

leolara (approved):

leolara:
Great!!

leolara:
@10to4 this was approved, you can merge.

PR #29, opened 2023-05-18
added less than gadget to stdlib

qwang98:
Requires `to_repr` function implemented by eth-types package of zkevm-circuits repo to work. Current implementation of this function by `FieldExt` type doesn't work.

To import eth-types, we need to update the `halo2_proofs` dependency to use the latest tag. It's currently using `v2023_02_02`. 

Otherwise let me know if my implementation is correct @leolara


qwang98:
@leolara I merged main over and now it's ready for review again :)

qwang98:
@leolara Merged main and fixed all formatting issues. 

Equivalent to this zkEVM `less_than` gadget: https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/gadgets/src/less_than.rs

leolara:
We will revisit this in the future

Issue #31, opened 2023-05-18
Benchmark Bytecode circuit from zkEVM PR using Plaff


qwang98:
The following are just my notes for Plaff and benching the bytecode circuit. Leo and Edu was supposed to work on most of it as discussed on May 9th but I'm also happy to help.

Plaff (Polyexen repo from Edu's alterego profile): https://github.com/Dhole/polyexen/commits?author=Dhole

## [LL] TODO: for current zkEVM PR

- define some criteria we all agree with at which point we can merge it
- review the gates that are generated by chiquito and make sure they are not higher degree than previous one and that they are similar
- another criteria could be the performance of witness assignment
- to export from Halo2, we can use plaff?
    - send compiled halo2 to plaff (though need to compile to the latest halo2 version we use in zkEVM)
    - edu can do that locally
    - Edu is still working on plaff
- **[SW]** benchmark witness generation (make sure we not on different order of magnitude)
    - just put a timer before and after running it in main
    - use without and with chiquito
- details on Plaff
    - depend on a third fork of Halo2 that Edu’s working on
    - need to create a branch from Leo’s PR and chain the Halo2 from Edu
    - alterego of Edu made it in a repo called polyexen
    - plaff is an intermediate representation in toml
    - can take circuit produced in Halo2 and output a toml
    - can use this as a tool to compare efficiency between halo2 and chiquito compiled halo2

qwang98:
documented my understandings here: https://mud-jaguar-906.notion.site/plaf-8ab8997389954d6289be588606d10719

qwang98:
Would also comment that using Plaf with zkEVM per se is much easier :)
Using it with Chiquito is a bit more complicated and we should work on releasing a version using 4_23_23 halo2_proofs.

leolara:
@ed255 @qwang98  I think a fundamental part of this is that FieldExt has been removed from halo2 and in zkevm circuits being subsituted by eth_types::Field. In chiquito we cannot depend on eth_types::Field as it is inside zkevm, and we need it for F::one().

leolara:
@ed255 @qwang98 I solved in the branch. Created new version of chiquito that uses the same halo2 than zkevm main. And Changed the uses of ::one() and ::zero() to the new ::ONE and ::ZERO

ed255:
[...]

PR #34, opened 2023-06-04
Steve/issue 33 plaf backend

qwang98:
# Summary
Created `chiquito2Plaf` function that converts columns, poly constraint, lookups, and q_enable/q_first/q_last from ir circuit to Plaf. Toml output (in top level directory) fully tested with fibonacci and mimc7 examples. Cannot generate csv output for fixed assignment because it's not stored in ir circuit struct.

# Caveats
1. Leo mentioned in issue 33 that imported halo2 columns should be terminated with error. However, I found that Halo2Advice can be converted to plaf except that the phase is missing, so I defaulted phase to 0. Halo2Fixed can be fully converted to plaf, so I don't think we need to terminate it. I'd also appreciate an explanation of phase.
2. ir circuit has as fixed_gen lambda field, which modifies the `FixedGenContextHalo2` struct from halo2 backend. We can technically convert this struct to the fixed assignment field of plaf. Let me know if we want a function for this in our plaf backend.
3. See **detailed analysis** below, but there are fields that only ir circuit has and that only plaf has and I made some intentional choices. most notably, I cannot convert fixed assignment lambda in chiquito ir circuit to fixed assigment in plaf. Chiquito also doesn't have copy constraints.

# Detailed analysis
## Chiquito ir circuit
```
pub struct Circuit<F, TraceArgs, StepArgs> {
    [NO NEED FOR PLAF] pub placement: Placement<F, StepArgs>, // Plaf doesn't concern placement, as query rotations are already included in both chiquito query and plaf query
    [NO NEED FOR PLAF] pub selector: StepSelector<F, StepArgs>, // Plaf doesn't concern step selector
    [NO NEED FOR PLAF] pub step_types: HashMap<u32, Rc<StepType<F, StepArgs>>>, // plaf doesn't concern step type

    [CONVERTED] pub q_enable: Column, // included as fixed column in plaf
    [CONVERTED] pub q_first: Option<Column>, // included as fixed column in plaf
    [CONVERTED] pub q_last: Option<Column>, // included as fixed column in plaf

    [CONVERTED] pub columns: Vec<Column>,
    [CONVERTED] pub polys: Vec<Poly<F>>,
    [CONVERTED] pub lookups: Vec<PolyLookup<F>>,

    [NO NEED FOR PLAF] pub trace: Option<Rc<Trace<TraceArgs, StepArgs>>>, // plaf doesn't concern
    [CANNOT CONVERT] pub fixed_gen: Option<Rc<FixedGen<F>>>, // fixed gen assign function implementation and assigned results in FixedGenContextHalo2 struct is in Chiquito's halo2 backend and not in IR; this field is a lambda, not sure how to convert the lamda to Plaf's Vec<Vec<_>> format for fixed assignments
}
```
## Plaf
```
pub struct Plaf {
    /// Circuit general information
    [NOT IN CHIQUITO] pub info: Info,
    /// Column information
    [CONVERTED] pub columns: Columns,
    /// List of polynomial identity constraints
    [CONVERTED] pub polys: Vec<Poly>,
    /// Matadata
    [NOT IN CHIQUITO] pub metadata: Metadata, // used for supporting halo2 query names in Plaf frontend, no corresponding fireld in chiquito
    /// List of lookup constraints
    [CONVERTED] pub lookups: Vec<Lookup>,
    /// List of copy constraints
    [NOT IN CHIQUITO] pub copys: Vec<CopyC>, // chiquito has no copy constraints
    /// Assignment values to the fixed columns.  None is used for non-assigned values, which means
    /// a 0 value.
    [CANNOT CONVERT] pub fixed: Vec<Vec<Option<BigUint>>>,
}
```


leolara:
@qwang98 i will make more comments but the files that are output of plaf should not be commited, usually we don't commit output. The example can produce the output but it is not commited. Perhaps you can include them in .Guti gnome

src/backend/plaf.rs
```diff
@@ -0,0 +1,285 @@
...
+                plaf.columns.witness.push(plaf_witness); 
+            }
+            cFixed => {
+                let plaf_fixed = ColumnFixed::new(column.annotation.clone());
+                self.add_id_index_mapping(column, c_column_id_to_p_column_index, fixed_index.unwrap());
+                plaf.columns.fixed.push(plaf_fixed);
+            }
+            Halo2Advice => { // ??? QUESTION: should terminate with error but only phase is missing from chiquito imported halo2 advice, so I defaulted phase to 0. is this good?
+                let plaf_witness = ColumnWitness::new(
+                    column.annotation.clone(),
```
leolara, line 154: When the user imports a column from Halo2 expects it to be the same column to share info, but you are creating a new column in plaf that is not the same as the Halo2 column. The only solution is to emit an error

src/backend/plaf.rs
```diff
@@ -0,0 +1,285 @@
...
+        ChiquitoPlaf {
+            debug,
+            circuit,
+            c_column_id_to_p_column_index: HashMap::new(),
+            advice_index: Counter::new(),
+            fixed_index: Counter::new(),
+        }
+    }
+
+    pub fn chiquito2Plaf(&mut self) -> Plaf {
```
leolara, line 61: The function chiquito2Plaf is a top level function like in the other backend. Inside the function you can create a struct and call methods in it.

src/backend/plaf.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub fn increment(&mut self) {
+        self.number += 1;
+    }
+
+    pub fn number(&self) -> usize {
+        self.number
+    }
+}
+
+pub fn write_files<F: PrimeField<Repr = [u8; 32]>, TraceArgs, StepArgs: Clone>(
```
leolara, line 265: I don't think we need to write files. Once we create the plaf object the user can use the plaf functions and utils

src/backend/plaf.rs
```diff
@@ -0,0 +1,285 @@
...
+    } 
+}
+
+
+#[derive(Clone, Debug)]
+pub struct Counter {
+    number: usize,
+}
+
+impl Counter {
```
leolara, line 251: I don't think we need to wrap an usize in a struct. We can just use the usize.

src/backend/plaf.rs
```diff
@@ -0,0 +1,285 @@
...
+
+    fn convert_plaf_query(
+        &self,
+        column: &cColumn,
+        rotation: &i32,
+        _annotation: &String,
+        index: usize, // plaf index starts from 0 for each column type
+    ) -> ColumnQuery {
+        match column.ctype {
+            cAdvice | Halo2Advice => {
```
leolara, line 223: @qwang98 for the same reason we cannot use the imported halo2 columns, we cannot also have queries to halo2 columns

leolara:
In general, the witness generation and the fixed generation is missing, but this needs more design so we should talk about this.

placement, selector, step_type and trace are used for witness generation.

fixed_gen is used for fixed generation.

But we cannot "insert" these into the plaf object, we will need to return other objects that can carry this operation while communicating with plaf.

src/backend/plaf.rs
qwang98: Fixed following the halo2 backend

src/backend/plaf.rs
qwang98: Fixed by emitting an error. Could you explain what you mean by "not the same as the Halo2 column"? My logic was that Plaf fixed column only requires a name string, which the imported halo2 fixed can supply, so I think we can port it from halo2 to plaf, but I might be missing something. Similar for advice column as well.

leolara: The import of columns is to share the column with another halo2 circuit. So the creates plaf column is not same as that halo2 column. Like in the bytecode circuit you share the column with the keccak circuit or push data circuit to lookup. If you create this new column it does not have the same witness or is actually the same.

qwang98: Ah this makes much more sense from a higher level. Thanks for the explanation!

qwang98: Also I used panic! for all error handling to follow the same practice in other Chiquito files. This is not the end game "proper" error handling but I'm assuming this is fine for now :)

src/backend/plaf.rs
qwang98: Fixed

src/backend/plaf.rs
qwang98: Fixed

src/backend/plaf.rs
qwang98: I commented this part out, as it's great for debugging purpose to see the toml and csv outputs and this function can be reused in many circuits. Let me know if you want me to delete this entirely to keep the code clean though!

examples/mimc7.rs
```diff
@@ -337,3 +338,20 @@ mod mimc7_constants {
         "13602139229813231349386885113156901793661719180900395818909719758150455500533",
     ];
 }
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
qwang98: Moved the test to main for MiMC7 as a proof of concept. It's currently entirely commented out as it won't work. Note that Plaf won't be generated for MiMC7, which imports fixed column from Halo2. More importantly, we cannot hook to any Plaf backend for now because we need to generate a `Witness` object as well defined in Plaf, which contains all the witness assignments.

The commented out Plaf boilerplate is still helpful, as it demonstrates a PoC how we can use a Plaf backend. Plaf only has Halo2 backend for now, which is not very helpful for now as Chiquito has a Halo2 backend already.

Removed the test entirely from Fibo. Will add again once we finish witness gen and fixed gen part of Chiquito Plaf backend.

qwang98:
> @qwang98 i will make more comments but the files that are output of plaf should not be commited, usually we don't commit output. The example can produce the output but it is not commited. Perhaps you can include them in .gitignore

All plaf output are now generated in the output folder of the top level directory. I also included output folder in .gitignore.

qwang98:
@leolara Addressed all comments. For the examples, we cannot generate any Plaf backend, as it requires a Witness object which contains all assignments. I wrote and commented out boilerplate codes as an incomplete PoC.

leolara:
Rustfmt and clippy @qwang98 

Screenshot 2023-06-06 at 13 57 26

qwang98:
# Plaf backend part II: fixed and witness assignment (WIP not ready for review)
I decided not to start another issue as it's easier with the merge main

# Fixed assignments
## Notes taken from Leo 
- `fixed_gen` from IR is used for fixed generation.
- new struct for plaf implementing fixedgencontext trait of chiquito
- call `fixed_gen` lambda with the new struct and put information into the vec<vec<fixed>> field of plaf
## Steve's additional thoughts
- we can adapt this new struct from a struct in chiquito backend called `FixedGenContextHalo2`, which the `fixed_gen` lambda manipulates
- `FixedGenContextHalo2` has a `Vec<Assignment<F, Fixed>>`

# Witness (advice) assignments
## Notes taken from Leo
- `placement`,` selector`, `step_type` and `trace` are used for witness generation. They are all from IR.
- But we cannot "insert" these into the plaf object, we will need to return other objects that can carry this operation while communicating with plaf.
## Steve's additional thoughts
###  IR `placement` field
This is a `Placement` object. 
```
pub struct Placement<F, StepArgs> {
    pub forward: HashMap<ForwardSignal, SignalPlacement>,
    pub steps: HashMap<Rc<StepType<F, StepArgs>>, StepPlacement>,
    pub columns: Vec<Column>,
}
```
Its most granular level is a `SignalPlacement` that contains an IR `Column` and rotation
```
pub struct SignalPlacement {
    pub column: Column,
    pub rotation: i32,
}
```
### IR `selector` field
This is a `StepSelector` object, which contains mappings of StepType to PolyExpr (selector). This is a one to one relationship, meaning each StepType can only have one selector_expr and one selector_expr_not and one selector_assigment

examples/fibonacci.rs
```diff
@@ -207,4 +210,40 @@ fn main() {
...
     }
+
+    // plaf boilerplate
+    use chiquito::backend::plaf::{ChiquitoPlaf, utils::alias_replace, chiquito2Plaf, print_witness};
+    use polyexen::plaf::{Witness, backends::halo2::PlafH2Circuit};
+
+    // get Chiquito ir
+    let circuit = fibo_circuit::<Fr>();
+    // get Plaf
+    let mut plaf_output = chiquito2Plaf(circuit, 8, false);
```
leolara, line 221: let (mut plaf, mut plaf_wit_gen) =

leolara:
A few methods have "halo2" in the name when they should be called "plaf"

leolara:
Remove commented out methods

leolara:
Comments should be written as sentences starting with capital and ending each sentece in dot.

leolara:
@qwang98 why are we changing the version of halo2_proofs? we should keep the same as in the zkEVM main.

examples/fibonacci.rs
qwang98: Fixed!

qwang98:
> A few methods have "halo2" in the name when they should be called "plaf"

All fixed.

> Remove commented out methods

All removed, including the `write_files` debugging method which I'll look for in commit history in the future (note to self).

> Comments should be written as sentences starting with capital and ending each sentece in dot.

All fixed.

> @qwang98 why are we changing the version of halo2_proofs? we should keep the same as in the zkEVM main.

We need the same version as Plaf to be able to use Plaf's functions. That said, we also need compatibility with zkEVM main. I think we should go with Edu's latest comment at the bottom of [this issue](https://github.com/privacy-scaling-explorations/chiquito/issues/31). This is supposed to achieve compatibility for both Plaf and zkEVM, but I'm confused why its Cargo.toml still use a different version from zkEVM for both halo2_proofs and its patch [here](https://github.com/ed255/polyexen-demo/blob/f73accf1e2a503f903ef50218f853f0d9a1c963e/Cargo.toml#L22-L25).

qwang98:
All comments addressed. Now ready for review with questions **bolded** below. @leolara 

# Witness Assignments
Plaf's has a few key differences from Chiquito in terms of witness assignments:
- In Plaf, A `k` is needed to create `2^k` rows for the PLONKish table. Number of rows is stored in `Plaf::Info.num_rows`. This is similar to regular Halo2.
- Chiquito has a `height` field for `TraceWitness` and uses `max_offset` for witness and fixed generation. These are used to prevent from assigning outside an offset value, though I do not see `set_height` used anywhere. I didn't use `max_offset` as Plaf doesn't have this field, but **should I also check that Plaf assignments happen within offset of `2^k`?**

# Fixed Assignments
FIxed assignments won't work, because currently Chiquito relies on imported Halo2 fixed. I wrote boilerplate code for Plaf fixed assignments and left TODOs that can be completed once we have a Chiquito native fixed column type. I think we should keep these code for now, because:
- They won't cause additional error, because Chiquito circuits with fixed columns (and thus imports from Halo2) will panic regardless.
- It's easier if we can review and merge these code as a template for now, so that we don't need to start from scratch in the future. **Let me know if you disagree.**

# Testing
I tested witness generation using the `print_witness` debugging method in Plaf backend on the Fibonacci circuit. It tests all functions though there's no `TraceArgs`.

# Miscellaneous
Not sure what to do with halo2_proof versioning in Cargo.toml. **Please review my prior comment.**

leolara:
@qwang98 please move the version of plaf one commit back and move halo2 to the right version and then let me know to do a final review.

examples/mimc7.rs
```diff
@@ -244,6 +246,48 @@ fn main() {
             println!("{}", failure);
         }
     }
+
+    // // plaf boilerplate
```
leolara, line 250: @qwang98 please removed unused code

src/backend/plaf/utils.rs
```diff
@@ -0,0 +1,67 @@
+use convert_case::{Case, Casing};
```
leolara, line 1: @qwang98 please, remove this I don't think we need it in principle

src/backend/plaf.rs
```diff
@@ -0,0 +1,507 @@
...
+
+#[derive(Clone, Debug)]
+pub struct ChiquitoPlaf<F: PrimeField, TraceArgs, StepArgs: Clone> {
+    debug: bool,
+    circuit: cCircuit<F, TraceArgs, StepArgs>,
+    // Chiquito column id doesn't start from zero.
+    // Plaf column index starts from 0 for each column type (advice, fixed, and instance).
+    // Therefore a mapping is needed to convert chiquito column id to plaf index.
+    c_column_id_to_p_column_index: HashMap<u32, usize>,
+    advice_index: usize, // This is Plaf's index of witness (advice) column.
```
leolara, line 53: do we use this variable for something? it seems it is assigned once but never used

src/backend/plaf.rs
```diff
@@ -0,0 +1,507 @@
...
+        let p = get_field_p::<F>();
+        plaf.info.p = p;
+
+        plaf.info.num_rows = 2usize.pow(k);
+
+        let mut c_column_id_to_p_column_index = HashMap::<u32, usize>::new();
+        let mut advice_index = 0;
+        let mut fixed_index = 0;
+
+        for column in self.circuit.columns.iter() {
```
leolara, line 88: circuit.columns is a vec and hence each column has an index, you can use that index as the plaf index and then you do not need all the c_column_id_to_p_column_index, advice_index and fixed_index stuff

src/backend/plaf.rs
```diff
@@ -0,0 +1,507 @@
...
+            if self.debug {
+                println!("MAP {:#?}", c_column_id_to_p_column_index);
+            }
+        }
+
+        self.c_column_id_to_p_column_index = c_column_id_to_p_column_index;
+        self.advice_index = advice_index;
+        self.fixed_index = fixed_index;
+
+        if !self.circuit.polys.is_empty() {
```
qwang98, line 108: Can delete the if as the loop won't run if it's empty

src/backend/plaf.rs
```diff
@@ -0,0 +1,507 @@
...
+                plaf.polys.push(plaf_poly);
+            }
+        }
+
+        for lookup in self.circuit.lookups.iter() {
+            let v1 = lookup
+                .exprs
+                .clone()
+                .into_iter()
+                .map(|(e1, _)| self.convert_plaf_poly(&e1))
```
qwang98, line 123: .map(|(e1, e2)| (self.convert_plaf_poly(&e1), self.convert_plaf_poly(&e2)))

qwang98: Better names for v1, v2

src/backend/plaf.rs
qwang98: Deleted entirely. Will live in commit history.

examples/mimc7.rs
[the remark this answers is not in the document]
qwang98: Deleted comments entirely and will live in commit history.

examples/mimc7.rs
qwang98: Fixed

src/backend/plaf/utils.rs
qwang98: Deleted from plaf/utils and Cargo.toml

src/backend/plaf.rs
qwang98: Fixed

src/backend/plaf.rs
qwang98: As discussed, this is still needed because Plaf has separate indices advice and fixed (they both start from zero).

src/backend/plaf.rs
qwang98: Fixed

src/backend/plaf.rs
qwang98: Fixed. As discussed, the above won't work as it's converting a vector of tuples to a tuple of vectors. Used `fold` instead.

qwang98:
@leolara  All comments addressed and halo2_proof versioning also fixed. Ready for final review before merging :)

.gitignore
```diff
@@ -1,2 +1,3 @@
 /target
 /Cargo.lock
+/output
```
leolara, line 3: what is writing to this directoty?

src/backend/plaf.rs
```diff
@@ -0,0 +1,474 @@
...
+            .c_column_id_to_p_column_index
+            .get(&placement.column.uuid())
+            .unwrap_or_else(|| panic!("plaf column not found for forward signal {:?}", forward));
+
+        (*p_column_index, super_rotation)
+    }
+}
+
+// This is for debugging only.
+pub fn print_witness(plaf_witness: &pWitness) {
```
leolara, line 469: @qwang98 no code is using this, remove it

src/backend/plaf/utils.rs
```diff
@@ -0,0 +1,65 @@
+use polyexen::plaf::Plaf;
+
+// adapted from polyexen-demo
+// keep a version for Chiquito plaf backend that we can maintain independently
+pub fn alias_replace(plaf: &mut Plaf) {
```
leolara, line 5: remove it

src/backend/plaf.rs
```diff
@@ -0,0 +1,474 @@
...
+            Halo2Advice | Halo2Fixed => {
+                panic!("Imported Halo2Advice and Halo2Fixed are not supported")
+            }
+        }
+    }
+}
+
+// ChiquitoPlafFixedGen currently doesn't work due to reasons mentioned in comments below.
+// We left skeleton code and TODOs for future implementations.
+pub struct ChiquitoPlafFixedGen {
```
leolara, line 259: not a good idea to leave half completed code that does not work, remove ChiquitoPlafFixedGen, just panic if the circuit has fixed_gen or Halo2FixedQuery queriables imported.

leolara: ok, the panic is already there, remove ChiquitoPlafFixedGen

.gitignore
leolara: if not of our code write into output, remove this line, the code in general should not have refences to thigns that you are doing only in local

qwang98: Done!

src/backend/plaf.rs
qwang98: Fibonacci example is actually using it as a demo.

src/backend/plaf/utils.rs
qwang98: Done. Was previously used for MiMC7 example boilerplate. The chunk has been deleted.

src/backend/plaf.rs
qwang98: I simply removed all fixed gen related codes. I don't think we should panic if there's fixed gen, because:
1. We already panic when there's imported halo2 fixed column.
2. It's still helpful to output the Plaf struct (without fixed gen) and Plaf Witness struct even if a Chiquito circuit has fixed gen.

Let me know if you disagree though.

qwang98:
All Leo's comments addressed or responded to. Final review :)

src/backend/plaf.rs
leolara: ok, so then move this code to the example. This is not part of the API of the backend, the backend outputs a plaf object, doing something with that plaf object is not part of the backend

leolara:
Just one thing left @qwang98

src/backend/plaf.rs
qwang98: Fixed

leolara (approved):

PR #42, opened 2023-06-14
Steve/issue 40 public signal new

qwang98:


src/compiler.rs
```diff
@@ -172,6 +174,13 @@ impl<CM: CellManager, SSB: StepSelectorBuilder> Compiler<CM, SSB> {
         unit.forward_signals = sc.forward_signals.clone();
 
         self.cell_manager.place(&mut unit);
+
+        for forward_signal in sc.exposed.clone() {
+            let forward_placement = unit.placement.get_forward_placement(&forward_signal);
+            let exposed: (Column, i32) = (forward_placement.column, forward_placement.rotation);
```
qwang98, line 180: don't need to type

src/dsl.rs
```diff
@@ -33,6 +33,15 @@ impl<F, TraceArgs, StepArgs> CircuitContext<F, TraceArgs, StepArgs> {
         Queriable::Forward(self.sc.add_forward(name, phase), false)
     }
 
+    pub fn expose(&mut self, forward: Queriable<F>) {
+        match forward {
+            Queriable::Forward(forward_signal, _) => {
+                self.sc.expose(forward_signal);
+            }
+            _ => panic!("Expected forward signal"),
```
qwang98, line 41: jarrrl

src/backend/halo2.rs
```diff
@@ -145,6 +148,23 @@ impl<F: Field + From<u64> + Hash, TraceArgs, StepArgs: Clone>
                 Ok(())
             },
         );
+
+        let mut instance_column_row: usize = 0;
+        for (column, rotation) in self.circuit.exposed.clone() {
```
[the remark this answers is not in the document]
qwang98: Fixed

src/compiler.rs
qwang98: Fixed!

qwang98:
Completed public signal backends for both Halo2 and Plaf. Ready for review @leolara. Note that I pr'ed this to the plaf backend branch, which you can review first.

src/backend/plaf.rs
```diff
@@ -121,6 +122,39 @@ impl<F: PrimeField<Repr = [u8; 32]>, TraceArgs, StepArgs: Clone>
...
+            // Public column not pulled from Chiquito ir, because it's not stored anywhere.
+            // Therefore, we create a Plaf public column from scratch.
+            let plaf_public = ColumnPublic::new(String::from(
+                "exposed forward signal values in first step instance",
+            ));
+            plaf.columns.public.push(plaf_public);
+        }
+
+        for (index, (c_column, rotation)) in self.circuit.exposed.iter().enumerate() {
+            let public_column = pColumn {
```
leolara, line 135: @qwang98 shouldn't this be the same as plaf_public?

src/backend/plaf.rs
```diff
@@ -121,6 +122,39 @@ impl<F: PrimeField<Repr = [u8; 32]>, TraceArgs, StepArgs: Clone>
...
+                kind: ColumnKind::Public,
+                index: 0, // Chiquito only has one public column, so the index is always 0.
+            };
+
+            let witness_index = self
+                .c_column_id_to_p_column_index
+                .get(&c_column.uuid())
+                .unwrap();
+
+            let witness_column = pColumn {
```
leolara, line 145: @qwang98 shouldn't we retrieve the actual column from plaf, instead of making a new one?

src/backend/plaf.rs
qwang98: They are actually different structs.

plaf_public is of `ColumnPublic` type, which is stored in the `Plaf` data struct.
```
pub struct ColumnPublic {
    pub name: String,
    pub aliases: Vec<String>,
}
```

public_column is of `Column` type, which is used for expressing copy constraints in this context.
```
pub struct Column {
    pub kind: ColumnKind,
    pub index: usize,
}
```

src/backend/plaf.rs
qwang98: As seen in the prior comment. The actual column stored in `Plaf` struct is different from `Column` used to express Plaf's copy constraints. The copy constraints also take ownership of the `Column` struct, and that's why I make a new one. Again, the `Column` struct here is really just used to express copy constraints and therefore duplicating them should be fine.

```
pub struct CopyC {
    pub columns: (expr::Column, expr::Column),
    pub offsets: Vec<(usize, usize)>,
}

pub struct Column {
    pub kind: ColumnKind,
    pub index: usize,
}
```

leolara:
@qwang98 why is this PR going to merge into steve/issue_33-plaf_backend instead of main? this is very confusing. I fork the branch to check something but somethings are not there

leolara:
I think the plaf backend is missing

qwang98:
> I think the plaf backend is missing

Apologies! I brought it back, together with your comments and my responses for it, if you expand this page.

leolara (approved):

PR #43, opened 2023-06-16
improve: use closure in add_lookup

dyxushuai:
Using closure in `add_lookup` can avoid exposing the details of `LookupBuilder`, and maybe will be good for lazy execution in the future.

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -3,7 +3,7 @@ use chiquito::{
     compiler::{
         cell_manager::SingleRowCellManager, step_selector::SimpleStepSelectorBuilder, Compiler,
     },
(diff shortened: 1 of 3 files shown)


leolara:
I don't see the advantage of this, I would close it. Closures causes moves of variables so I would only use them when there is no better alternative.

leolara:
@dyxushuai If you are interested in chiquito, we would love to talk with you, where can I find you?

dyxushuai:
@leolara Of course, I currently reside in Shanghai/China at UTC+8 timezone. You can touch me in my e-mail address, johnxu@duck.com.
If you guys have any chat groups or other communication platforms, please let me know.

PR #45, opened 2023-06-22
steve/issue_44-lookup_table_abstraction

qwang98:
# Syntax
## New lookup table 
On the circuit context level:
```
let table = lookup_table(vec![dest_col_1, dest_col_2, ...]);
```
## Match lookup table
On the step type setup context level:
```
ctx.add_lookup(table.match(vec![source_col_1, source_col_2]).enable(enabler));
```
OR
```
ctx.add_lookup(table.enable(enabler).match(vec![source_col_1, source_col_2]));
```

# Code Structure
- Reorganized all lookup related contents and APIs to `lb.rs`, which stands for "lookup builder", although there's no LookupBuilder struct any more.
- Moved `Lookup` struct and APIs from ast to dsl::lb, because users need to call `match` and `enable` on `Lookup` in some scenarios, and user functions must be in dsl. 
- ast still has its own `Lookup` struct, but I moved all APIs to dsl::lb.
- An alternative is to define a LookupBuilder {lookup_table, lookup} and bring all APIs under that.

# TODO
Please refer to the MiMC7 example for an issue with my comments. Basically the step type def closure takes ownership of the abstracted `LookupTable` even though `match` and `enable` only take a reference. I'm not sure how to solve this and tried shared reference using `Arc` but the syntax doesn't work well for our dsl.

One solution could be bringing back the lookup builder, which takes a reference to the table and then builds the `Lookup`, so we have a builder for each lookup. The syntax will be more complicated and looks like:
```
ctx.add_lookup(lookup(table).enable(enabler).match(vec![source_col_1, source_col_2]));
```
It doesn't look too bad though, but let me know if you have a better solution.


qwang98:
Also please ignore the API doc comments for the functions. I haven't updated them and plan to do that once the code is more stable.

qwang98:
# Design Requirements
- Keep `Lookup` in ast
- Keep prior APIs for `LookupBuilder` and methods for building lookup from (src_col, dest_col) tuples
- Add APIs for building lookup using `LookupTable`
- Use a handler for `LookupTable`, because Copy trait cannot be implemented for it

# Prior API
In dsl, LookupBuilder modifies a Lookup, by calling `add(src_col, dest_col)` and `enable(e)`. Both functions return `&mut LookupBuilder` for nested calls.
```
pub struct LookupBuilder<F> {
    pub lookup: Lookup<F>,
}

pub fn add<C: Into<Constraint<F>>, E: Into<Expr<F>>>(
    &mut self,
    constraint: C,
    expression: E,
) -> &mut Self

pub fn enable<C: Into<Constraint<F>>>(&mut self, enable: C) -> &mut Self
```
On step type level, `add_lookup` adds Lookup in LookupBuilder to ast::StepType:
```
pub fn add_lookup(&mut self, lookup_builder: &mut LookupBuilder<F>)
```
`Lookup` and methods for modifying it are all in ast.

# Proposal
Because `LookupTable` only contains destination columns, we need another struct that contains the `Lookup` we are building.  We need to return the final `Lookup` from this struct and not from `LookupTable`. We also cannot directly call ast functions to build `Lookup`. Therefore, I think we still need a `LookupBuilder` that contains a `LookupTableHandler`, rather than having `LookupTable` and `LookupBuilderGeneric` both implement a `LookupBuilder` trait, because we cannot return `Lookup` from a `LookupTable` that only contains destination columns.

qwang98:
Outdated. See draft pr #47.

PR #49, opened 2023-06-26
issue 48 part 1: dsl and ast for shared signal

qwang98:
Currently compiler errors out, which will be fixed in another PR.

### DIFF
--- a/src/dsl.rs
+++ b/src/dsl.rs
@@ -33,6 +33,10 @@ impl<F, TraceArgs, StepArgs> CircuitContext<F, TraceArgs, StepArgs> {
         Queriable::Forward(self.sc.add_forward(name, phase), false)
     }
 
+    pub fn shared(&mut self, name: &str) -> Queriable<F> {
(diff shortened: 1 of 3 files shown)


qwang98:
Clippy and fmt will be fixed in the final part of the shared signal issue, not in this PR.

src/dsl.rs
```diff
@@ -33,6 +33,10 @@ impl<F, TraceArgs, StepArgs> CircuitContext<F, TraceArgs, StepArgs> {
         Queriable::Forward(self.sc.add_forward(name, phase), false)
     }
 
+    pub fn shared(&mut self, name: &str) -> Queriable<F> {
```
leolara, line 36: Please add shared_with_phase to the DSL API

leolara (approved):

src/dsl.rs
qwang98: I will merge this for now and fix this in Part 2, so that I only need to sync this change to Part 3.

PR #50, opened 2023-06-26
issue 48 part 2: compiler and ir 

qwang98:
PR'ed against another branch (issue 48 part 1).

IR didn't need to change, because all signals are compiled into ir::Column and Placement, which is defined in compiler/cell manager.

### DIFF
--- a/src/compiler/cell_manager.rs
+++ b/src/compiler/cell_manager.rs
@@ -1,5 +1,5 @@
 use core::panic;
-use std::{collections::HashMap, fmt::Debug, ops::Div, rc::Rc};
+use std::{collections::HashMap, fmt::Debug, rc::Rc};
 
 use crate::ast::{ForwardSignal, InternalSignal, SharedSignal, StepType};
 
@@ -122,7 +122,6 @@ impl CellManager for SingleRowCellManager {
             forward_signals += 1;
         }
 
-        // SingleRowCellManager has height 1 for all steps and thus works for SharedSignal.
         let mut shared_signals: u32 = 0;
 
         for shared_signal in unit.shared_signals.iter() {
@@ -197,22 +196,8 @@ pub struct MaxWidthCellManager {
 
 impl CellManager for MaxWidthCellManager {
     fn place<F, StepArgs>(&self, unit: &mut CompilationUnit<F, StepArgs>) {
-        // If there exists shared signals, determine whether MaxWidthCellManager will return steps
-        // with different heights. If yes, panic.
-        // Can also use `same_height` method after placing all signals.
-        // However, checking at the front is more efficient by skipping placement if steps don't
-        // have the same height.
         if !unit.shared_signals.is_empty() {
-            let max_width = self.max_width;
-            let mut height_by_step_type = unit.step_types.values().map(|step_type| {
-                (step_type.signals.len() + unit.forward_signals.len() + unit.shared_signals.len())
-                    .div(max_width)
-                    + 1
-            });
-            let first = height_by_step_type.next().unwrap();
-            if !height_by_step_type.all(|len| len == first) {
-                panic!("To use shared signal, all steps placed by MaxWidthCellManager need to have the same height. Please use a different cell manager.");
-            }
+            panic!("Shared signals are not supported for MaxWidthCellManager, which might return steps with variable heights.");
         }
 
         let mut placement = Placement::<F, StepArgs> {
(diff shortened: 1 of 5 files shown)


src/compiler/cell_manager.rs
```diff
@@ -163,8 +197,27 @@ pub struct MaxWidthCellManager {
...
+        // have the same height.
+        if !unit.shared_signals.is_empty() {
+            let max_width = self.max_width;
+            let mut signal_len_by_step_type = unit.step_types.values().map(|step_type| {
+                (step_type.signals.len() + unit.forward_signals.len() + unit.shared_signals.len())
+                    .div(max_width)
+                    + 1
+            });
+            let first = signal_len_by_step_type.next().unwrap();
+            if !signal_len_by_step_type.all(|len| len == first) {
```
qwang98, line 213: `signal_len_by_step_type` should be `step_height_by_step_type`. Sorry for the confusion.

src/compiler/cell_manager.rs
```diff
@@ -249,6 +334,11 @@ impl CellManager for MaxWidthCellManager {
             placement.steps.insert(Rc::clone(step), step_placement);
         }
 
+        // Redundant if we do the step height check at the front.
+        // if !placement.same_height() {
+        //     panic!("To use shared signal, all steps placed by MaxWidthCellManager need to have
+        // the same height. Please use a different cell manager."); }
```
qwang98, line 340: Let me know if you want me to do this check at the front (so that I delete these comments) or at the back (so that I uncomment these).

src/dsl.rs
```diff
@@ -33,10 +33,18 @@ impl<F, TraceArgs, StepArgs> CircuitContext<F, TraceArgs, StepArgs> {
...
+    /// `Queriable` instance representing the added shared signal.
     pub fn shared(&mut self, name: &str) -> Queriable<F> {
         Queriable::Shared(self.sc.add_shared(name, 0), 0)
     }
 
+    /// Adds a shared signal to the circuit with a name string and a specified phase and returns a
+    /// `Queriable` instance representing the added shared signal.
+    pub fn shared_with_phase(&mut self, name: &str, phase: usize) -> Queriable<F> {
+        Queriable::Shared(self.sc.add_shared(name, phase), 0)
+    }
```
qwang98, line 46: Made these changes in this PR, so that I only need to merge in Part 3.

qwang98:
Not sure why CI is not running, but I ran clippy and rustfmt. They can be fixed in the last part though.

src/dsl.rs
leolara: I don't understand this sentence. Why is this showing this up if this is from part 1.

src/compiler/cell_manager.rs
```diff
@@ -163,8 +197,27 @@ pub struct MaxWidthCellManager {
 
 impl CellManager for MaxWidthCellManager {
     fn place<F, StepArgs>(&self, unit: &mut CompilationUnit<F, StepArgs>) {
+        // If there exists shared signals, determine whether MaxWidthCellManager will return steps
```
leolara, line 200: So for this CM specifically I think it is better to panic always if there are shared columns. We could revisit this in the future

leolara (changes requested):
A quick review, 

After placement, the compiler should check that all steps are the same height if shared are not empty, we cannot leave that to the Cell Manager implementer, that could be someone else.

Change MaxWidthCellManager to always panic if there are shared signals.

src/dsl.rs
qwang98: Apologies! I merged part 1 to main before fixing this so that I can merge main in part 2 rather than asking you to approve changes in part 1 again. This way it's quicker. However, I think you are right that for code organization reason it's the best to keep it in part 1.

src/compiler/cell_manager.rs
qwang98: Fixed. Removed associated code for shared columns and panic at the top if there are shared columns.

qwang98:
> A quick review,
> 
> After placement, the compiler should check that all steps are the same height if shared are not empty, we cannot leave that to the Cell Manager implementer, that could be someone else.
> 
> Change MaxWidthCellManager to always panic if there are shared signals.

Fixed!

qwang98:
Ready for review again.

leolara (approved):

PR #53, opened 2023-06-27
issue 52: fixed signal part 1

qwang98:
DSL + AST

Note that fixed signal has no phase.

Again, clippy and fmt will be fixed in the last part of the pull request.


leolara (approved):

leolara:
there is a lot of CI that don't pass

PR #54, opened 2023-06-28
issue 52 part 2: ir, compiler, and halo2 backend for fixed gen

qwang98:
IR, compiler, and halo2 backend

Note that assignment won't work with rotation (`next`, `rot(i32)`, and `prev`), because FixedGenContext `assign` function takes three parameters: `offset` (absolute row number), `lhs` (Queriable) and `rhs` (value).

Rotation for fixed signal should work for constraints though.

I also deleted shared signal codes from `MaxWidthCellManager` because it panics at the very front. Should have caught this in the shared signal PR.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -16,6 +16,7 @@ use halo2_proofs::plonk::{Advice, Column as Halo2Column, ColumnType, Fixed};
 pub struct Circuit<F, TraceArgs, StepArgs> {
     pub forward_signals: Vec<ForwardSignal>,
     pub shared_signals: Vec<SharedSignal>,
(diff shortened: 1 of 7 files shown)


leolara:
For the assignment you cannot allow queriables that have rotation. But you need to use the rotation in the placement, and do placement.rotation + (offset * step_height)

qwang98:
Ready for review again. Sorry for the delay as I just arrived at home and rested a bit today.

qwang98:
> For the assignment you cannot allow queriables that have rotation. But you need to use the rotation in the placement, and do placement.rotation + (offset * step_height)

Fixed

leolara (approved):

PR #56, opened 2023-07-01
Refactor for SuperCircuit

leolara:


src/wit_gen.rs
```diff
@@ -20,63 +22,68 @@ impl<F> StepInstance<F> {
...
-                .get(&step.uuid())
-                .expect("step type not found"),
-        );
-
+impl<F> TraceContext<F> {
+    pub fn add<Args, WG: Fn(&mut StepInstance<F>, Args) + 'static>(
+        &mut self,
+        step: &StepTypeWGHandler<F, Args, WG>,
+        args: Args,
+    ) {
```
qwang98, line 57: What's the purpose of having `StepTypeWGHandler` and storing `wg` lambda there instead of storing `wg` in `StepType`?

src/compiler.rs
```diff
@@ -275,28 +255,26 @@ impl<CM: CellManager, SSB: StepSelectorBuilder> Compiler<CM, SSB> {
...
+                lookups: unit.lookups,
+                step_types: unit.step_types,
+                q_enable,
+                q_first,
+                q_last,
+
+                fixed_gen: sc.fixed_gen.as_ref().map(|v| Rc::clone(v)),
+            },
+            sc.trace.as_ref().map(|v| TraceGenerator::new(Rc::clone(v))),
+        )
```
qwang98, line 280: What's the purpose of separating out `trace` while previously it's a part of the ir::Circuit?

src/wit_gen.rs
leolara: That is the whole point. StepType and Circuit and ir::Circuit should not be parametrized by StepArgs.

src/compiler.rs
leolara: As before

src/wit_gen.rs
```diff
@@ -20,63 +22,68 @@ impl<F> StepInstance<F> {
...
+impl<F: Default, TraceArgs> TraceGenerator<F, TraceArgs> {
+    pub fn new(trace: Rc<Trace<F, TraceArgs>>) -> Self {
+        Self { trace }
+    }
+
+    pub fn generate(&self, args: TraceArgs) -> TraceWitness<F> {
+        let mut ctx = TraceContext::default();
+
+        (self.trace)(&mut ctx, args);
+
```
qwang98, line 88: What's the purpose of calling `trace` in wit_gen instead of in the backend?

src/wit_gen.rs
```diff
@@ -87,3 +89,48 @@ impl<F: Default, TraceArgs> TraceGenerator<F, TraceArgs> {
...
+        } else {
+            let mut assigments = vec![F::ZERO; self.num_steps];
+            assigments[offset] = rhs;
+            self.assigments.insert(lhs, assigments);
+        }
+    }
+
+    pub fn get_assiggnments(self) -> FixedAssignment<F> {
+        self.assigments
+    }
```
qwang98, line 128: Spelling for "assignment".

src/wit_gen.rs
```diff
@@ -87,3 +89,48 @@ impl<F: Default, TraceArgs> TraceGenerator<F, TraceArgs> {
...
+            num_steps,
+        }
+    }
+
+    /// Takes a `Queriable` object representing the fixed column (lhs) and the value (rhs) to be
+    /// assigned.
+    pub fn assign(&mut self, offset: usize, lhs: Queriable<F>, rhs: F) {
+        if !Self::is_fixed_queriable(lhs) {
+            panic!("trying to assign not fixed signal");
+        }
```
qwang98, line 115: Typo. Say "Trying to assign non-fixed signal" instead.

src/compiler.rs
```diff
@@ -174,12 +174,16 @@ impl<CM: CellManager, SSB: StepSelectorBuilder> Compiler<CM, SSB> {
...
+        if !unit.placement.same_height() {
+            panic!("Cannot calculate the number of rows");
         }
 
+        unit.num_rows = ast.num_steps * (unit.placement.first_step_height() as usize);
+
+        self.compile_fixed(ast, &mut unit);
+
+        self.compile_exposed(ast, &mut unit);
+
```
qwang98, line 186: We are limiting ALL circuits to those with the same step height for all step instances?

src/compiler.rs
```diff
@@ -492,6 +525,48 @@ impl<CM: CellManager, SSB: StepSelectorBuilder> Compiler<CM, SSB> {
...
+                column_values[offset] = value;
+                offset += step_height as usize;
+            }
+
+            result.insert(placement.column, column_values);
+        }
+
+        result
+    }
+
```
qwang98, line 569: Why don't we do this in CellManager?

src/compiler.rs
```diff
@@ -332,6 +338,33 @@ impl<CM: CellManager, SSB: StepSelectorBuilder> Compiler<CM, SSB> {
...
+        &self,
+        ast: &astCircuit<F, TraceArgs>,
+        unit: &mut CompilationUnit<F>,
+    ) {
+        for forward_signal in ast.exposed.clone() {
+            let forward_placement = unit.placement.get_forward_placement(&forward_signal);
+            let exposed = (forward_placement.column, forward_placement.rotation);
+            unit.exposed.push(exposed);
+        }
+    }
```
qwang98, line 351: Note to self: I think we should add `expose` for shared signal as well.

src/compiler.rs
```diff
@@ -492,6 +525,48 @@ impl<CM: CellManager, SSB: StepSelectorBuilder> Compiler<CM, SSB> {
...
+                column_values.clone()
+            } else {
+                empty.clone()
+            };
+
+            let mut offset = placement.rotation as usize;
+            for value in assignments {
+                column_values[offset] = value;
+                offset += step_height as usize;
+            }
```
qwang98, line 562: Looks like we limit the total number of fixed rows to the number of step instances. This isn't necessarily true, especially when we have large lookup tables that contain fixed columns. Any reason for this design?

It also looks like fixed column assignments start from step instance zero and increment by 1 for each assignment. I'm not sure what's the purpose of tying step instance index to fixed assignments? Also if a step's height is greater than one, we only assign one row for the fixed column, which can be a waste of space.

qwang98:
Looks like we do wit_gen and call trace in `TraceGenerator::generate`, which is called separately. wit gen is then passed into backend `synthesize` as a parameter. 

However, we call fixed_gen in the compiler and store fixed assignments in `ir::Circuit`.

I noticed that Plaf has a similar design, i.e. fixed assignments in the circuit, while witness assignments are separate. What's the reason for this design?

src/compiler.rs
leolara: Yes, but later

qwang98 (approved):
All set except typos! Merging earlier will be great so I can expose Python bindings.

leolara:
Close in favor of https://github.com/privacy-scaling-explorations/chiquito/pull/57

For some reason this PR was not getting new commits

PR #58, opened 2023-07-08
Common code for assignments and default fixed moved to compiler

leolara:


.gitignore
```diff
@@ -1,2 +1,3 @@
 /target
 /Cargo.lock
+.vscode
```
qwang98, line 3: Space at the end :)

examples/fibonacci.rs
```diff
@@ -109,6 +108,8 @@ fn fibo_circuit<F: Field + From<u64> + Hash>() -> (Circuit<F>, Option<TraceGener
             })
         });
 
+        ctx.pragma_num_steps(11);
+
```
qwang98, line 112: Probably should have mentioned in the last PR, but looks like `num_steps` in AST is used to calculate `num_rows` in CompilationUnit, which is used to calculate the number of assignment rows. This is different from `first_step` and `last_step`, which create a fixed column eventually.

My question is: How does `num_steps` enforce the number of steps if not through a fixed column? Can the user still pass in a witness with different number of steps while still passing the circuit constraints?

leolara: https://github.com/privacy-scaling-explorations/chiquito/pull/58/files#diff-19a5e6a0406f9184fd74dba07991bee321e608450fff0e30117f02c49c97ea8aR585-R586

.gitignore
leolara: Done

src/wit_gen.rs
```diff
@@ -100,10 +100,10 @@ pub struct FixedGenContext<F> {
 }
 
 impl<F: Field + Hash> FixedGenContext<F> {
-    pub fn new(num_steps: usize) -> Self {
+    pub fn new(num_rows: usize) -> Self {
         Self {
             assigments: Default::default(),
-            num_steps,
+            num_steps: num_rows,
```
qwang98, line 106: Should we just call this field `num_rows` instead of `num_steps`?

src/compiler.rs
```diff
@@ -107,11 +110,54 @@ impl<F> CompilationUnit<F> {
...
+            fixed_signals: ast.fixed_signals.clone(),
+            ..Default::default()
+        }
+    }
+}
+
+pub struct CompilerConfig<CM: CellManager, SSB: StepSelectorBuilder> {
+    cell_manager: CM,
+    step_selector_builder: SSB,
+}
```
qwang98, line 136: What's the use of `CompilerConfig` as it's literally the same as `Compiler`. I only see it used once in super circuit.

src/wit_gen.rs
leolara: Done

src/compiler.rs
```diff
@@ -240,43 +230,32 @@ impl<CM: CellManager, SSB: StepSelectorBuilder> Compiler<CM, SSB> {
...
+                AssigmentGenerator::new(
+                    unit.columns,
+                    unit.placement,
+                    unit.selector,
+                    TraceGenerator::new(Rc::clone(v)),
+                    unit.num_rows,
+                    uuid,
+                )
+            }),
         )
```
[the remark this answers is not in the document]
leolara: step_types will never be needed, once we are in the paradigm of (ir::Column,rot) and not in the one of Queriables, all that staff is not needed, as it is needed to compile from Queriables to (ir::Column,rot)

qwang98:
Very excited about the super circuits! Just to check my understanding, an example will be:

super_circuit: SuperCircuit = super_circuit("super circuit", |ctx: SuperCircuitContext| {
  ctx.add(compiler_config, ast_circuit_1);
  ctx.add(compiler_config, ast_circuit_2);
  ctx.mapping(|ctx: MappingContext, (u64, F, bool)| {
    ctx.map(assignment_generator_1, (u64, F)); // assignment generators are returned from compiler.compile
    ctx.map(assignment_generator_2, (F, bool));
  }
}

mapping_generator = MappingGenerator { super_circuit.mapping }
super_assignments = mapping_generator.generate((u64, F, bool));

Then, in halo2 boilerplate, call `chiquito2Halo2` on each of the ir::Circuit objects in the super circuit.

After that, in the `synthesize` function, loop through each of the `ChiquitoHalo2` objects, and call halo2::synthesize on the corresponding assignment object from super_assignments with the same uuid.

My question is: Will super circuit implement the Circuit trait from Halo2 and its `without_witness`, `configure`, `synthesize` functions?

src/compiler.rs
leolara: It is used here: https://github.com/privacy-scaling-explorations/chiquito/pull/58/files#diff-477cf4d3133e8c73e0a761339f1d361e646943fe2f8f5c332ccdf96df759ed66R31

I think in the end the compiler will not need to contain this fields, they will be removed.

leolara:
> Very excited about the super circuits! Just to check my understanding, an example will be:
> 
> ```
> super_circuit: SuperCircuit = super_circuit("super circuit", |ctx: SuperCircuitContext| {
>   ctx.add(compiler_config, ast_circuit_1);
>   ctx.add(compiler_config, ast_circuit_2);
>   ctx.mapping(|ctx: MappingContext, (u64, F, bool)| {
>     ctx.map(assignment_generator_1, (u64, F)); // assignment generators are returned from compiler.compile
>     ctx.map(assignment_generator_2, (F, bool));
>   }
> }
> 
> mapping_generator = MappingGenerator { super_circuit.mapping }
> super_assignments = mapping_generator.generate((u64, F, bool));
> ```
> 
> Then, in halo2 boilerplate, call `chiquito2Halo2` on each of the ir::Circuit objects in the super circuit.
> 
> After that, in the `synthesize` function, loop through each of the `ChiquitoHalo2` objects, and call halo2::synthesize on the corresponding assignment object from super_assignments with the same uuid.
> 
> My question is: Will super circuit implement the Circuit trait from Halo2 and its `without_witness`, `configure`, `synthesize` functions?

So, in the case of halo2, for each sub_circuit there will be a different `region`.

We will need to add stuff to the halo2 backend, and some boilerplate from the examples will be moved there.

Also, we will add the possibility of adding raw halo2 sub circuits to the super circuit, but perhaps we can leave that for the future when someone needs it.

leolara:
@qwang98 please, go again on it

examples/fibonacci.rs
qwang98: Then I'd just comment that we are only allowing circuits with the same height for all steps, which I think we discussed last time and are fine with it for now.

src/compiler.rs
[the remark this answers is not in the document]
qwang98: Do you mean `place_queriable` which converts `Queriable` to `PolyExpr::Query(column, rot, annotation)`?

src/compiler.rs
qwang98: Kk. I saw it's used in the link. Just saying that it's the same struct as Compiler basically. Good with eventually removing them! :)

qwang98:
> > Very excited about the super circuits! Just to check my understanding, an example will be:
> > ```
> > super_circuit: SuperCircuit = super_circuit("super circuit", |ctx: SuperCircuitContext| {
> >   ctx.add(compiler_config, ast_circuit_1);
> >   ctx.add(compiler_config, ast_circuit_2);
> >   ctx.mapping(|ctx: MappingContext, (u64, F, bool)| {
> >     ctx.map(assignment_generator_1, (u64, F)); // assignment generators are returned from compiler.compile
> >     ctx.map(assignment_generator_2, (F, bool));
> >   }
> > }
> > 
> > mapping_generator = MappingGenerator { super_circuit.mapping }
> > super_assignments = mapping_generator.generate((u64, F, bool));
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then, in halo2 boilerplate, call `chiquito2Halo2` on each of the ir::Circuit objects in the super circuit.
> > After that, in the `synthesize` function, loop through each of the `ChiquitoHalo2` objects, and call halo2::synthesize on the corresponding assignment object from super_assignments with the same uuid.
> > My question is: Will super circuit implement the Circuit trait from Halo2 and its `without_witness`, `configure`, `synthesize` functions?
> 
> So, in the case of halo2, for each sub_circuit there will be a different `region`.
> 
> We will need to add stuff to the halo2 backend, and some boilerplate from the examples will be moved there.
> 
> Also, we will add the possibility of adding raw halo2 sub circuits to the super circuit, but perhaps we can leave that for the future when someone needs it.

Got it! Might do it as a future PR then. :)

qwang98:
I'd just comment on the fact that fixed columns created through `fixed_gen` have the same length as the number of steps rather than the number of rows (step * step height). Just to make sure this is intended.

leolara:
> I'd just comment on the fact that fixed columns created through `fixed_gen` have the same length as the number of steps rather than the number of rows (step * step height), which we used previously. Just to make sure this is intended. Otherwise, I'm good with everything and feel free to merge.

Yes, so I changed it to num_steps

qwang98 (approved):
All good!

PR #63, opened 2023-07-14
Super Circuit backend for halo2

leolara:
Also, includes common boilerplate for halo2

### DIFF
--- a/src/backend/halo2.rs
+++ b/src/backend/halo2.rs
@@ -2,10 +2,10 @@ use std::{collections::HashMap, hash::Hash};
 
 use halo2_proofs::{
     arithmetic::Field,
-    circuit::{Cell, Layouter, Region, RegionIndex, Value},
+    circuit::{Cell, Layouter, Region, RegionIndex, SimpleFloorPlanner, Value},
     plonk::{
-        Advice, Any, Column, ConstraintSystem, Error, Expression, FirstPhase, Fixed, Instance,
-        SecondPhase, ThirdPhase, VirtualCells,
+        Advice, Any, Circuit as h2Circuit, Column, ConstraintSystem, Error, Expression, FirstPhase,
+        Fixed, Instance, SecondPhase, ThirdPhase, VirtualCells,
     },
     poly::Rotation,
 };
@@ -14,6 +14,7 @@ use crate::{
     ast::ToField,
     ir::{
         assigments::Assignments,
+        sc::SuperAssignments,
         Circuit, Column as cColumn,
         ColumnType::{Advice as cAdvice, Fixed as cFixed, Halo2Advice, Halo2Fixed},
         PolyExpr,
@@ -26,7 +27,7 @@ pub fn chiquito2Halo2<F: Field + From<u64> + Hash>(circuit: Circuit<F>) -> Chiqu
     ChiquitoHalo2::new(circuit)
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub struct ChiquitoHalo2<F: Field + From<u64>> {
     pub debug: bool,
 
@@ -35,16 +36,20 @@ pub struct ChiquitoHalo2<F: Field + From<u64>> {
     advice_columns: HashMap<UUID, Column<Advice>>,
     fixed_columns: HashMap<UUID, Column<Fixed>>,
     instance_column: Option<Column<Instance>>,
+
+    ir_id: UUID,
 }
 
 impl<F: Field + From<u64> + Hash> ChiquitoHalo2<F> {
     pub fn new(circuit: Circuit<F>) -> ChiquitoHalo2<F> {
+        let ir_id = circuit.id;
         ChiquitoHalo2 {
             debug: true,
             circuit,
             advice_columns: Default::default(),
             fixed_columns: Default::default(),
             instance_column: Default::default(),
+            ir_id,
         }
     }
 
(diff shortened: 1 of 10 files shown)


src/backend/halo2.rs
```diff
@@ -305,3 +310,110 @@ pub fn to_halo2_advice<F: Field>(
...
+    }
+
+    fn configure_with_params(
+        meta: &mut ConstraintSystem<F>,
+        mut compiled: Self::Params,
+    ) -> Self::Config {
+        compiled.configure(meta);
+
+        compiled
+    }
```
qwang98, line 348: We are using configure with params instead of configure. Is it to force the function to take ChiquitoHalo2 objects? A bit confused by Halo2's doc for `configure_wiht_params`:
```
    /// The circuit is given an opportunity to describe the exact gate
    /// arrangement, column arrangement, etc.  Takes a runtime parameter.  The default
    /// implementation calls `configure` ignoring the `_params` argument in order to easily support
    /// circuits that don't use configuration parameters.
```

leolara: We need to use configure_wiht_params in order to be able to have the same struct represent all chiquito circutis, as normal configure does not accept parameters, we could not give the chiquito circuit to it

src/compiler/linker.rs
```diff
@@ -0,0 +1,9 @@
+use crate::ir::{sc::SuperCircuit, Circuit};
+
+pub struct Linker {}
+
+impl Linker {
+    pub fn link<F>(units: Vec<Circuit<F>>) -> SuperCircuit<F> {
+        let sc = SuperCircuit::default();
+    }
+}
```
qwang98, line 9: This file isn't linked to the module tree. Need to `pub mod linker` in compiler.rs.

qwang98: I see that it's probably not completed yet, so that's why we aren't adding it to the module tree.

src/ir/assigments.rs
```diff
@@ -28,6 +27,19 @@ pub struct AssigmentGenerator<F, TraceArgs> {
...
+            columns: self.columns.clone(),
+            placement: self.placement.clone(),
+            selector: self.selector.clone(),
+            trace_gen: self.trace_gen.clone(),
+            num_rows: self.num_rows,
+            ir_id: self.ir_id,
+        }
+    }
+}
+
```
qwang98, line 42: How is it different from the default implementation? `usize` and `UUID` should be both copied rather than cloned in this implementation, but that isn't much different. I probably am missing something?

src/wit_gen.rs
```diff
@@ -71,11 +71,18 @@ impl<F> TraceContext<F> {
...
 }
 
+impl<F, TraceArgs> Clone for TraceGenerator<F, TraceArgs> {
+    fn clone(&self) -> Self {
+        Self {
+            trace: self.trace.clone(),
+        }
+    }
+}
+
```
qwang98, line 85: Same question.

src/compiler/linker.rs
leolara: I will remove this file.

src/ir/assigments.rs
leolara: The derived implementation requires TraceArgs to be Clone, when it is not necessary.

src/wit_gen.rs
leolara: The derived implementation requires TraceArgs to be Clone, when it is not necessary.

src/compiler/linker.rs
leolara: Done

src/compiler/step_selector.rs
```diff
@@ -14,14 +14,25 @@ use super::{Column, CompilationUnit, PolyExpr};
...
+    fn clone(&self) -> Self {
+        Self {
+            selector_expr: self.selector_expr.clone(),
+            selector_expr_not: self.selector_expr_not.clone(),
+            selector_assignment: self.selector_assignment.clone(),
+            columns: self.columns.clone(),
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
leolara: This is really not justified, I am removing it

src/backend/halo2.rs
```diff
@@ -305,3 +310,110 @@ pub fn to_halo2_advice<F: Field>(
...
+
+    fn synthesize(
+        &self,
+        sub_circuits: Self::Config,
+        mut layouter: impl Layouter<F>,
+    ) -> Result<(), Error> {
+        for sub_circuit in sub_circuits {
+            sub_circuit.synthesize(&mut layouter, self.witness.get(&sub_circuit.ir_id))
+        }
```
qwang98, line 411: If we get witness using ir_id, does it mean the SuperAssignments HashMap can only have one assignment for each ir_id? What about circumstances where we want to plugin multiple witnesses for the same subcircuit? Do we plug in a different SuperAssignments entirely?

src/backend/halo2.rs
qwang98: Got it! Thanks.

src/wit_gen.rs
qwang98: Got it. A bit surprised that you could work around the requirement like this. Learned something new today. ;)

src/compiler/step_selector.rs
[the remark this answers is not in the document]
leolara: Done

qwang98:
Just to clear my understanding, the linker is not needed, is it because we can simply link them via lookup tables? Lookup tables will be on the super circuit level and can be shared around different sub circuits and that's how we link the sub circuits? Will these lookup tables be global static and on the same level as the super circuit?

qwang98:
To clear my understanding, on a high level, we will generate SuperCircuit specified in ir/sc.rs, and generate super assignments using MappingGenerator. From there, we can construct the ChiquitoHalo2SuperCircuit in halo2 backend, and finally generate/verify proofs?

leolara:
> To clear my understanding, on a high level, we will generate SuperCircuit specified in ir/sc.rs, and generate super assignments using MappingGenerator. From there, we can construct the ChiquitoHalo2SuperCircuit in halo2 backend, and finally generate/verify proofs?

Yes

Issue #62, opened 2023-07-14
Allow to expose signals in arbitrary step offset


leolara:
To the ExposeOffset enum add a new value `Step(usize)` and then implement to expose that singal at that step instance.

qwang98:
This is done in #65.

Issue #61, opened 2023-07-14
Enable to expose signals for the last step instance


leolara:
Currently you can expose signals only for the first step instance with:

`ctx.expose(signal)`

Add an enum `ExposeOffset` with value `First` and  `Last`, so it can be used as second parameter:

`ctx.expose(signal, offset)`

qwang98:
This is done in #65.

PR #65, opened 2023-07-14
Feat/expose sigs and offset 60 62

### LINKED ISSUE #60: Enable to expose shared signals
Currently, the DSL function expose only accepts forward signals, add the possibility for shared signals.

lancenonce:
Fixed #60 , #61 , and #62 .

Need more clarity on issue 62; how is the offset expressed in the circuit? Is it the `i32` value in `ir::Circuit`'s `pub exposed: Vec<(Column, i32)>` ? Logic should be straight-forward to implement in `expose_first`, `expose_last`, and `expose_step` in `ast::Circuit`.


src/ast.rs
```diff
@@ -12,6 +12,27 @@ pub use expr::*;
 
 use halo2_proofs::plonk::{Advice, Column as Halo2Column, ColumnType, Fixed};
 
+use self::query::Queriable;
+
+/// Signal abstraction
+#[derive(Clone, Debug)]
+pub enum Signal {
```
leolara, line 19: I don't think we should have a new enum just for the exposed feature, just check in ast::expose could recieve a Queriable, and then panic for when no forward or shared.

Fixed signal cannot be exposed.

src/dsl.rs
```diff
@@ -53,14 +53,29 @@ impl<F, TraceArgs> CircuitContext<F, TraceArgs> {
     }
 
     /// Exposes the first step instance value of a forward signal as public.
-    pub fn expose(&mut self, forward: Queriable<F>) {
-        match forward {
+    pub fn expose(&mut self, queriable: Queriable<F>, offset: ExposeOffset) {
+        match queriable {
             Queriable::Forward(forward_signal, _) => {
-                self.sc.expose(forward_signal);
+                let signal = Signal::Forward(forward_signal);
```
leolara, line 59: @lancenonce I think it is better to pass Queriable to ast::expose, so we don't need to add an enum just for the expose feature

src/ast.rs
```diff
@@ -173,6 +194,24 @@ impl<F, TraceArgs> Circuit<F, TraceArgs> {
 
         Rc::clone(step_rc)
     }
+
+    pub fn expose_first(&mut self, signal: Signal) {
```
leolara, line 198: In the AST you just need to store the information, then the compiler will convert it to column, rotation

leolara: So you can store with the  ExposeOffset enum. And perhaps you don't need three methods then.

leolara:
@lancenonce there is no logic in the ast, it is just a container with information. You should just store the signal, offset to be exposed. Then the compiler should convert that in the IR to (Column,rotation).

The backend is already taking the (Column,rotation) so it should work from there without modification.

lancenonce:
Thank you for these helpful comments @leolara ! I'm on it.

lancenonce:
Updates:

- We no longer use a `Signal` enum, just a Queriable that checks if the exposed signals are Forward / Shared
- We compile the exposed signals by returning the column and rotation depending on the signal and the offset passed from `ast`
- We store exposed signals in ast::Circuit with a `Vec<(Queriable<F>, ExposeOffset)>`

src/dsl.rs
```diff
@@ -162,6 +162,13 @@ pub enum StepTypeDefInput {
     String(&'static str),
 }
 
+#[derive(Clone, Copy)]
+pub enum ExposeOffset {
```
leolara, line 166: I think this enum should be in the ast.rs module

src/compiler.rs
```diff
@@ -325,12 +326,38 @@ impl<CM: CellManager, SSB: StepSelectorBuilder> Compiler<CM, SSB> {
...
     ) {
-        for forward_signal in ast.exposed.clone() {
-            let forward_placement = unit.placement.get_forward_placement(&forward_signal);
-            let exposed = (forward_placement.column, forward_placement.rotation);
+        for (queriable, offset) in &ast.exposed {
+            let exposed = match queriable {
+                Queriable::Forward(forward_signal, _) => {
+                    let placement = unit.placement.get_forward_placement(forward_signal);
+                    match offset {
+                        ExposeOffset::First => (placement.column, 0),
```
leolara, line 334: So if First should be equivalent to step=0 and Last to step=num_steps - 1

leolara (changes requested):

src/compiler.rs
```diff
@@ -328,9 +328,39 @@ impl<CM: CellManager, SSB: StepSelectorBuilder> Compiler<CM, SSB> {
...
+                            let rot = placement.rotation + ((unit.num_steps - 1) as i32) * (unit.placement.first_step_height() as i32);
+                            (placement.column, rot)
+                        },
+                        ExposeOffset::Step(step) => {
+                            let rot = placement.rotation + (*step as i32) * (unit.placement.first_step_height() as i32);
+                            (placement.column, rot)
+                        }
+                    }
+                }
+                Queriable::Shared(shared_signal, _) => {
```
leolara, line 347: @lancenonce the shared signal is still with the wrong values for rotation

lancenonce: Fixed, thanks @leolara !

src/frontend/pychiquito.rs
```diff
@@ -68,7 +68,7 @@ impl<'de> Visitor<'de> for CircuitVisitor {
                     if exposed.is_some() {
                         return Err(de::Error::duplicate_field("exposed"));
                     }
-                    exposed = Some(map.next_value::<Vec<ForwardSignal>>()?);
+                    exposed = Some(map.next_value::<Vec<(Queriable<Fr>, ExposeOffset)>>()?);
```
lancenonce, line 71: @qwang98 this is where the breaking change lies. Queriables do not implement any serde traits.

leolara (approved):

src/frontend/pychiquito.rs
qwang98: I merged over this branch and I'm working on this :)

src/ast.rs
```diff
@@ -106,9 +108,18 @@ impl<F, TraceArgs> Circuit<F, TraceArgs> {
...
+            Queriable::Forward(..) | Queriable::Shared(..) => {
+                self.exposed.push((signal, offset));
+            }
+            _ => {
+                // TODO: Add a proper error here
+                panic!("cannot expose fixed or internal signals")
+            }
+        }
     }
+
```
qwang98, line 122: Checking for Forward and Shared should already be done in the dsl `expose` function, so here we can simply push the `(signal, offset)`.

src/dsl.rs
```diff
@@ -53,12 +53,12 @@ impl<F, TraceArgs> CircuitContext<F, TraceArgs> {
...
-            Queriable::Forward(forward_signal, _) => {
-                self.sc.expose(forward_signal);
+    pub fn expose(&mut self, queriable: Queriable<F>, offset: ExposeOffset) {
+        match queriable {
+            Queriable::Forward(..) | Queriable::Shared(..) => {
+                self.sc.exposed.push((queriable, offset));
             }
-            _ => panic!("Expected forward signal"),
+            _ => panic!("Expected either ForwardSignal or SharedSignal"),
         }
```
qwang98, line 62: Here in `expose` we should check for Forward and Share as you've done, and then call `expose` in the AST rather than directly pushing to `exposed` of AST.

In general, we want functions that modifies AST to remain in AST, so that DSL can do simple calls of AST functions. Let me know if this makes sense!

leolara (dismissed):
Waiting for Steve comments to be done

src/dsl.rs
leolara: @qwang98 good catch

src/ast.rs
leolara: I think it is better to do it in AST, if we are going to do it only in one place. As if someone modifies the AST without the DSL the check should be done

lancenonce:
Changes should be good to go, please let me know what you think of the new Error enum, I'm thinking we can use this for more precise error handling @qwang98 @leolara

lancenonce:
@qwang98 to merge pychiquito fixes after this is merged in.

qwang98:
> Changes should be good to go, please let me know what you think of the new Error enum, I'm thinking we can use this for more precise error handling @qwang98 @leolara

New fixes look good to me!

Re error handling, we've been panic'ing everything so far, which is not the proper way lol. @leolara I think we'd eventually need precise error handling everywhere. What do you think?

qwang98:
I fixed frontend related stuffs and should be ready for merge once @leolara approves again.

leolara:
> > Changes should be good to go, please let me know what you think of the new Error enum, I'm thinking we can use this for more precise error handling @qwang98 @leolara
> 
> New fixes look good to me!
> 
> Re error handling, we've been panic'ing everything so far, which is not the proper way lol. @leolara I think we'd eventually need precise error handling everywhere. What do you think?

Panicking is good if the user is creating a chiquito program that is not compilable

leolara:
> I fixed frontend related stuffs and should be ready for merge once @leolara approves again.

@qwang98 you are now the the reviewer, it can be merged when you approve it

qwang98:
> > > Changes should be good to go, please let me know what you think of the new Error enum, I'm thinking we can use this for more precise error handling @qwang98 @leolara
> > 
> > 
> > New fixes look good to me!
> > Re error handling, we've been panic'ing everything so far, which is not the proper way lol. @leolara I think we'd eventually need precise error handling everywhere. What do you think?
> 
> Panicking is good if the user is creating a chiquito program that is not compilable

kk I reverted back to prior panic message just to be consistent throughout the program. We can think about error handling in the future. I think a systematic approach should still be good.

qwang98 (approved):

PR #66, opened 2023-07-18
Enable that with MaxWidth CM is possible to force same height

leolara:


src/compiler/cell_manager.rs
```diff
@@ -330,6 +332,20 @@ impl CellManager for MaxWidthCellManager {
             placement.steps.insert(step.uuid(), step_placement);
         }
 
+        if self.same_height {
+            let height = placement
+                .steps
+                .iter()
+                .map(|(_, step)| step.height)
+                .max()
+                .unwrap_or(0);
```
qwang98, line 341: How about doing `expect` here to panic with an error message instead of `unwrap_or`?

qwang98: Just saying height = 0 doesn't make much sense in this context.

qwang98:
Also Clippy. Good with everything else!

src/compiler/cell_manager.rs
leolara: This will return None, ony if there is no steps, in a circuit without steps, 0 step height makes sense. Perhaps we can revisit this if we have circuits with only fixed signals.

qwang98: kk sounds good to me!

qwang98 (approved):

PR #67, opened 2023-07-19
Steve/frontend

qwang98:
For PyChiquito. Deserializes a json standard.


leolara (approved):
Only thing is to change the module to frontend/pychiquito.rs

PR #71, opened 2023-07-21
Feat/managed padding 68*

### LINKED ISSUE #68: Managed padding
In the trace context we can spec the padding step with its parameters:

```
    ctx.padding(step_type_hander, || step_args);
```

When doing witness generation, after finishing if there are more rows to fill according to the established height, then the padding step should be added until the circuit is filled.

This is specd in the trace function because `step_args` could depends on `trace_args`.

This does not need to be added to the AST or IR as it is included in the trace lambda. It should be included in the witness generation code.

The second parameter is a lambda to allow step_args to be different for each padding step instance, giving more generality.

lancenonce:
- Fixes #68 
- Adds padding functionality to `TraceContext`
- Fixes `Assignment` spelling
- Adds test for new padding function


leolara (dismissed):

src/wit_gen.rs
```diff
@@ -67,6 +71,29 @@ impl<F> TraceContext<F> {
...
+    // This is an external function users can use to create their own padding constraint vs. the default
+    pub fn set_padding<Args, WG: Fn(&mut StepInstance<F>, Args) + 'static>(&mut self, step_handler: &StepTypeHandler, lambda: PaddingLambda<F>) {
+        self.padding = Some((step_handler.uuid(), lambda));
+    }
+
+    // this function pads the rest of the circuit with the 
+    // StepInstance of the StepType in TraceContext::padding
+    fn pad(&mut self) {
+        // here we set the padding
+        while self.witness.step_instances.len() < self.witness.height {
```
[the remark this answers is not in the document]
lancenonce: This is added locally

src/wit_gen.rs
```diff
@@ -67,6 +71,29 @@ impl<F> TraceContext<F> {
...
+            if let Some((padding_uuid, padding_lambda)) = &self.padding {
+                let mut padded_witness = StepInstance::new(*padding_uuid);
+
+                // How should these TraceArgs be used? How do we add them to this scope without changing the type parameters for TraceContext?
+                let trace_args = padding_lambda();
+
+                self.witness.step_instances.push(padded_witness);
+            } else {
+                panic!("Missing padding step!");
```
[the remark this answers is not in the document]
lancenonce: Is an Ok() fine?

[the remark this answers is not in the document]
leolara: I dont think we need to return anything from pad method

src/wit_gen.rs
```diff
@@ -62,6 +66,27 @@ impl<F> TraceContext<F> {
 
         self.witness.step_instances.push(witness);
     }
+
+    pub fn set_num_steps(&mut self, num_steps: usize) {
```
leolara, line 70: @lancenonce we should not have this method, num_steps should be passed to the constructor of TraceContext

leolara:
Have you tested this?

src/wit_gen.rs
```diff
@@ -62,6 +62,21 @@ impl<F> TraceContext<F> {
...
+    // This function pads the rest of the circuit with the given StepTypeWGHandler
+    pub fn padding<Args, WG: Fn(&mut StepInstance<F>, Args) + 'static>(
+        &mut self,
+        step: &StepTypeWGHandler<F, Args, WG>,
+        args: Args,
+    ) {
+        let mut witness = StepInstance::new(step.uuid());
+
+        (*step.wg)(&mut witness, args);
+        let len = self.witness.step_instances.len();
```
leolara, line 75: As we talk before several times, the loop should be up to num_steps

src/wit_gen.rs
```diff
@@ -62,6 +62,21 @@ impl<F> TraceContext<F> {
...
+        for _ in 0..len {
```
leolara, line 76: This loop from 0 to current length will duplicate the number of step instances. Not bring it to num_steps

src/wit_gen.rs
```diff
@@ -62,6 +62,21 @@ impl<F> TraceContext<F> {
 
         self.witness.step_instances.push(witness);
     }
+
+    // This function pads the rest of the circuit with the given StepTypeWGHandler
+    pub fn padding<Args, WG: Fn(&mut StepInstance<F>, Args) + 'static>(
+        &mut self,
+        step: &StepTypeWGHandler<F, Args, WG>,
+        args: Args,
```
leolara, line 70: As we spoke several times args must be a lambda

leolara:
So to summarise:
 1. This task add a new method to TraceContext, called padding.
 2. This method have two parameters step: &StepTypeWGHandler<F, Args, WG>, argsFn: Fn(Args),
 3. The method will add step instances to the trace until the total is num_steps.
 4. For each step instance added, argsFn is called to get the Args that is pass to self.add.
 5. num_steps is set in TraceContext in the constructor when created. TraceContext::new(num_step).
 6. Where ever TraceContext is created, if num_steps is not available there, it will be injected there also in the constructor.

lancenonce:
> Have you tested this?

Tests done

src/wit_gen.rs
```diff
@@ -124,20 +148,61 @@ impl<F: Field + Hash> FixedGenContext<F> {
...
+
+    fn dummy_args_fn() {}
+
+    #[test]
+    fn test_padding_no_witness() {
+        let mut ctx = TraceContext::new(5);
+        let step = StepTypeWGHandler::new(
+            uuid(),
+            "dummy",
+            |_: &mut StepInstance<i32>, _: ()| {}, // replace i32 with your Field type
```
leolara, line 182: I don't think the comment is useful. Nobody should come and replace that in this test

src/wit_gen.rs
```diff
@@ -42,15 +42,25 @@ pub struct TraceWitness<F> {
...
+        }
+    }
+
     pub fn get_witness(self) -> TraceWitness<F> {
         self.witness
     }
 }
 
-impl<F> TraceContext<F> {
+impl<F: Clone> TraceContext<F> {
```
leolara, line 63: Is f: clone required anymore?

src/wit_gen.rs
```diff
@@ -42,15 +42,25 @@ pub struct TraceWitness<F> {
 #[derive(Debug, Default)]
 pub struct TraceContext<F> {
     witness: TraceWitness<F>,
+    num_steps: usize,
 }
 
 impl<F> TraceContext<F> {
+    pub fn new(num_steps: usize) -> Self {
+        Self {
+            witness: TraceWitness {
```
leolara, line 51: I think it is better to use witness: Default::default()

src/dsl.rs
```diff
@@ -356,6 +356,15 @@ pub struct StepTypeWGHandler<F, Args, D: Fn(&mut StepInstance<F>, Args) + 'stati
 }
 
 impl<F, Args, D: Fn(&mut StepInstance<F>, Args) + 'static> StepTypeWGHandler<F, Args, D> {
+    pub fn new(id: UUID, annotation: &'static str, wg: D) -> Self {
```
leolara, line 359: How was this struct created before?

src/wit_gen.rs
lancenonce: no, that was for a feature I tried and forgot to remove.

src/dsl.rs
lancenonce: with the `step_type_def()` function, but for testing adding a `new` method made it so that a new circuit wouldn't need to be created. Let me know if this shouldn't be here.

src/wit_gen.rs
```diff
@@ -9,7 +9,7 @@ use crate::{
 
 /// A struct that represents a witness generation context. It provides an interface for assigning
 /// values to witness columns in a circuit.
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone)]
```
leolara, line 12: @lancenonce I think this is the last thing, we should remove Default here, because we don't want people creating this with num_steps=0 by mistake.

leolara:
@lancenonce just one comment, the rest looks good :+1:

src/wit_gen.rs
leolara: So we are still waiting for this

lancenonce: Just committed the change, msgd you a few questions about this but this should be correct.

leolara: Thanks, I just realised that my comment was in the wrong line, but you did the right thing.

qwang98 (approved):
Looks all good to me!

PR #72, opened 2023-07-22
Working super circuit and LookupTable abstraction

leolara:


src/dsl/lb.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    /// Adds a selector column specific to the lookup table. Because the function returns a mutable
+    /// reference to the `LookupBuilder<F>`, it can an chain multiple `add` and `enable` function
+    /// calls to build the lookup table. Requires calling `lookup` to create an
+    /// empty `LookupBuilder` instance at the very front.
+    pub fn when<C: Into<Constraint<F>>>(mut self, enable: C) -> Self {
+        self.enable = Some(enable.into());
+
+        self
+    }
```
qwang98, line 190: Minor improvement: Check if self.enable is already Some, panic "cannot call `when` twice on the same LookupTableBuilder".

src/dsl/lb.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// Adds a selector column specific to the lookup table. Because the function returns a mutable
+    /// reference to the `LookupBuilder<F>`, it can an chain multiple `add` and `enable` function
+    /// calls to build the lookup table. Requires calling `lookup` to create an
+    /// empty `LookupBuilder` instance at the very front.
+    pub fn enable<C: Into<Constraint<F>>>(mut self, enable: C) -> Self {
+        let enable = enable.into();
+        self.lookup.enable(enable.annotation, enable.expr);
+        self
+    }
+}
```
qwang98, line 62: This looks like the deprecated lookup builder. It doesn't implement the `LookupBuilder` trait which `add_lookup` in dsl requires. It also doesn't have a public `new` method so it cannot really be created. Maybe we can just delete it?

src/dsl/lb.rs
```diff
@@ -0,0 +1,199 @@
...
+    pub fn add(&self, table: LookupTableStore<F>) {
+        self.0.lock().as_mut().unwrap().insert(table.uuid(), table);
+    }
+}
+
+impl<F: Clone> LookupTableRegistry<F> {
+    pub fn get(&self, uuid: UUID) -> LookupTableStore<F> {
+        (*self.0.lock().unwrap().get(&uuid).unwrap()).clone()
+    }
+}
```
qwang98, line 139: Just to clear my understanding, Arc<Mutex<>> is how we create a persistent variable for the super circuit? By persistent I mean it's not globally static, but will persist as long as we have one reference to it?
1. Are we using Arc instead of Rc so that it's thread-safe, i.e. multiple sub-circuits (threads) can access the same LookupRegistry?
2. Why use Mutex instead of RefCells?
3. Why is lazy static not needed? Is it because Arc will persist as long as there are more than one reference, so that we don't need a globally static variable?

qwang98: I'm thinking about using a similar design for PyChiquito Rust end, which stores HashMap<UUID, AST>, where UUID is shared across Python and Rust.

src/dsl/lb.rs
```diff
@@ -0,0 +1,199 @@
...
+        if let Some(enable) = enable {
+            lookup.enable(enable.annotation, enable.expr);
+        }
+
+        src.into_iter()
+            .zip(self.dest)
+            .for_each(|(src, dest)| lookup.add(src.annotation, src.expr, dest));
+
+        lookup
+    }
```
qwang98, line 111: Will we ever build directly from a LookupTableStore instead of from a LookupBuilder using an UUID? If no, we can make this function private.

src/dsl/sc.rs
```diff
@@ -1,57 +1,109 @@
...
+        (assignment, exports)
     }
 
     pub fn mapping<D: Fn(&mut MappingContext<F>, MappingArgs) + 'static>(&mut self, def: D) {
-        self.sc.set_mapping(def);
+        self.super_circuit.set_mapping(def);
+    }
+
+    fn compile(mut self) -> SuperCircuit<F, MappingArgs> {
+        // TODO: This is a hack, tbe compiler should be stateless
```
qwang98, line 71: Typo: "the" instead of "tbe"

src/dsl/sc.rs
```diff
@@ -1,57 +1,109 @@
...
     },
 };
 
+use super::{lb::LookupTableRegistry, CircuitContext};
+
 pub struct SuperCircuitContext<F, MappingArgs> {
-    sc: SuperCircuit<F, MappingArgs>,
+    super_circuit: SuperCircuit<F, MappingArgs>,
+    sub_circuit_phase1: Vec<CompilationUnit<F>>,
+    tables: LookupTableRegistry<F>,
```
qwang98, line 24: Looks like LookupTableRegistry is cloned from the super circuit to the circuit, and eventually to the step type. Does it mean adding a lookup table at the circuit level doesn't update it at the super circuit level?

If this is an issue, we could modify the `sub_circuit` function in dsl/sc.rs to add lookup tables from circuit lookuptableregistry to super circuit lookuptableregistry. Alternatively, we could simply delete the tables field from SuperCircuitContext if it's not needed.

src/ir/sc.rs
```diff
@@ -11,14 +11,14 @@ use super::{
 
 pub struct SuperCircuit<F, MappingArgs> {
     sub_circuits: Vec<Circuit<F>>,
-    mapping: Option<Box<Mapping<F, MappingArgs>>>,
+    mapping: MappingGenerator<F, MappingArgs>,
 }
```
qwang98, line 15: Just curious and for my learning but why are we changing from Box of Mapping to MappingGenerator struct which contains an Rc of Mapping? Is it because mapping now needs multiple ownerships? Why does it need multiple ownerships now?

src/ir/sc.rs
```diff
@@ -63,10 +75,30 @@ impl<F: Field> MappingContext<F> {
...
+    mapping: Rc<Mapping<F, MappingArgs>>,
+}
+
+impl<F, MappingArgs> Clone for MappingGenerator<F, MappingArgs> {
+    fn clone(&self) -> Self {
+        Self {
+            mapping: self.mapping.clone(),
+        }
+    }
+}
```
qwang98, line 87: Can be derived?

src/ir/assigments.rs
```diff
@@ -40,6 +40,19 @@ impl<F: Clone, TraceArgs> Clone for AssigmentGenerator<F, TraceArgs> {
...
+        Self {
+            columns: Default::default(),
+            placement: Default::default(),
+            selector: Default::default(),
+            trace_gen: Default::default(),
+            num_rows: Default::default(),
+            ir_id: Default::default(),
+        }
+    }
+}
```
qwang98, line 54: All use default. Can be derived then?

src/compiler/cell_manager.rs
```diff
@@ -84,7 +77,11 @@ impl Placement {
     }
 
     pub fn first_step_height(&self) -> u32 {
-        self.steps.values().next().unwrap().height
+        if let Some(step) = self.steps.values().next() {
+            step.height
+        } else {
+            self.base_height
+        }
     }
```
qwang98, line 85: Is this for circuits with no step, e.g. lookup table only circuits? Trying to envision how base_height is useful when there's no step.

src/compiler.rs
```diff
@@ -63,13 +74,24 @@ impl<F> Default for CompilationUnit<F> {
...
             fixed_assignments: Default::default(),
+
+            ast_id: Default::default(),
+            uuid: uuid(),
+
+            other_sub_circuits: Default::default(),
+            other_columns: Default::default(),
         }
     }
 }
```
qwang98, line 97: I think we can do:
```
...
uuid: uuid(),
..Default::default()
}
```

src/dsl/sc.rs
qwang98: What do you mean "stateless" compiler here? There's only one compile for the super circuit, so it should share the same state throughout the compile?

src/dsl/lb.rs
lancenonce: For 1), I don't think Rc is thread-safe, so it makes sense to use Arc if we are using multiple threads for the LookupRegistry. 

For 2), I think we need mutexes because the data in the LookupTableRegistry is being updated. I'm not sure about the breadth of their differences, but I know mutexes are usefule for safely managing threads that update state.

For 3), I am not sure, since I can see how using `lazy_static!` could make sense.

Of course, need confirmation from @leolara on these thoughts :)

qwang98:
To clear my understanding why we need to break `compile` into phase1 and phase2:
1. `compile_phase1` is called by `sub_circuit` to generate compiled sub circuit and assignment generator for the sub circuit.
2. `sub_circuit` also generates exports, such as lookup table, for other `sub_circuit` calls to use.
3. For each sub circuit, an assignment generator generates assignments via the `map` function.
4. `super_circuit` then calls `compile_phase2` on each phase1 compiled circuits to compile steps and add selector columns.

My guess is that, `compile_phase2` compiles constraints and lookups, which depend on `compile_phase1` outputs such as lookup tables, and that's why we need to break them up for two compile phases.

My questions are:
1. Why do we generate assignments between phase1 and phase2. Does phase2 depend on assignments?
2. Why we need to compile selector columns in phase2? Do they depend on exports or assignments?

src/dsl/lb.rs
leolara: It is an easy way to make a shared object writable (Rc only allows reads), and has the aditional benefit of being multi-thread, although we are not running it multithread now.

src/dsl/lb.rs
leolara: Done

src/dsl/sc.rs
leolara: Because it is made of `Arc<Mutex` the clone works as a shared reference

src/ir/sc.rs
leolara: Rc is better when you need to clone because it does not actually clone the element inside. I am not sure if we need the share reference really, but it is harder to find out.

src/ir/sc.rs
leolara: For deriving it, it requires F and MappingArgs to be Clone, which is not really necessary.

src/ir/assigments.rs
leolara: To be derived it requires F and TraceArgs to be Default, which is not really necessary.

src/compiler/cell_manager.rs
leolara: Yes, circuits with no steps like the mimc7 constant circuit. It will also help for circuits with steps of different height.

src/compiler.rs
leolara: We are implementing Default for this type, so if the compiler allows what you said it would end up in a recursive infinite loop.

leolara:
> To clear my understanding why we need to break `compile` into phase1 and phase2:
> 
> 1. `compile_phase1` is called by `sub_circuit` to generate compiled sub circuit and assignment generator for the sub circuit.
> 2. `sub_circuit` also generates exports, such as lookup table, for other `sub_circuit` calls to use.
> 3. For each sub circuit, an assignment generator generates assignments via the `map` function.
> 4. `super_circuit` then calls `compile_phase2` on each phase1 compiled circuits to compile steps and add selector columns.
> 
> My guess is that, `compile_phase2` compiles constraints and lookups, which depend on `compile_phase1` outputs such as lookup tables, and that's why we need to break them up for two compile phases.
> 
> My questions are:
> 
> 1. Why do we generate assignments between phase1 and phase2. Does phase2 depend on assignments?
> 2. Why we need to compile selector columns in phase2? Do they depend on exports or assignments?

1. Because SuperCircuitContext.sub_circuit needs to return the assignment so they can be capture by the mapping lambda.
2. Perhaps there are things that can be in any of the phases, It is harder if I have to analyse everything. The important thing is that things that require to use the placements of other sub_circuits have to be in phase two.

src/backend/halo2.rs
```diff
@@ -396,7 +416,26 @@ impl<F: Field + From<u64> + Hash> h2Circuit<F> for ChiquitoHalo2SuperCircuit<F>
...
+            sub_circuits.iter().fold(HashMap::default(), |mut acc, s| {
+                acc.extend(s.fixed_columns.clone());
+                acc
+            });
+
+        sub_circuits.iter_mut().for_each(|sub_circuit| {
+            sub_circuit.advice_columns = advice_columns.clone();
+            sub_circuit.fixed_columns = fixed_columns.clone();
+            sub_circuit.configure_sub_circuit(meta)
+        });
```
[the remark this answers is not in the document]
leolara: > Wondering what are the use cases of this,

The use case is SuperCircuits and lookuptables shared between sub_circuits.

> s almost everything is constrained within a sub circuit

A lookup from a imported lookup table contain expressions that query columns in another sub_circuit.

src/dsl/lb.rs
leolara: lazy_static is for global things, this is only at the super_circuit level

src/dsl/sc.rs
leolara: Fixed typo.

Compiler does not need to use &self in the future, because it does not make sense to keep state between compilations.

src/dsl/lb.rs
leolara: Done

src/dsl/lb.rs
qwang98: Got it thanks both!

src/dsl/sc.rs
qwang98: Got it!

src/dsl/sc.rs
qwang98: Ahh this is clever. Got it now.

src/ir/sc.rs
qwang98: I see.

qwang98 (approved):
I'm good with everything now! Maybe fix the clippy and fmt but those are minor, which can be done in the future.

Issue #83, opened 2023-08-16
Instance data in halo2


leolara:
We need to pass the instance (public column) to mock prover: https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/pi_circuit.rs#L1333-L1339

We should add a method to ChiquitoHalo2 struct called instance(&Assignments<F>)  -> Vec<F> that returns the values for the instance column from the assignements based on circuit.exposed translation to (column, offset).

Then ChiquitoHalo2Circuit should have the same method that just proxies to the contained ChiquitoHalo2, and wraps it in a vector with one element.

In ChiquitoHalo2SuperCircuit we add the same method, but we panic if there is more than one sub_circuit that returns something (we need to figure out in the future how to know the order of the instance columns in the instance vector).

leolara:
Peerhaps for the super circuit we can assume the order of the instance columns are the same as they are created with meta.instance_column() ; then we don't need to panic

PR #84, opened 2023-08-16
Updated PyChiquito frontend for latest changes

qwang98:
Most specifically, need to acquire num_step from ast to calculate padding in PyChiquito.

### DIFF
--- a/src/plonkish/backend/halo2.rs
+++ b/src/plonkish/backend/halo2.rs
@@ -170,7 +170,7 @@ impl<F: Field + From<u64> + Hash> ChiquitoHalo2<F> {
                 halo2_column,
                 // For single row cell manager, forward signal rotation is always zero.
                 // For max width cell manager, rotation can be non-zero.
-                // Offset is 0 + rotation for the first step instance.
+                // Offset is absolute row index calculated in `compile_exposed`.
                 *rotation as usize,
             );
             let _ = layouter.constrain_instance(cell, self.instance_column.unwrap(), index);
@@ -199,6 +199,25 @@ impl<F: Field + From<u64> + Hash> ChiquitoHalo2<F> {
         }
 
         Ok(())
+    }
+
+    fn instance(&self, witness: &Assignments<F>) -> Vec<F> {
+        let mut instance_values = Vec::new();
+        for (column, rotation) in &self.circuit.exposed {
+            let values = witness
+                .get(column)
+                .unwrap_or_else(|| panic!("exposed column not found: {}", column.annotation));
+
+            if let Some(value) = values.get(*rotation as usize) {
+                instance_values.push(*value);
+            } else {
+                panic!(
+                    "assignment index out of bounds for column: {}",
+                    column.annotation
+                );
+            }
+        }
+        instance_values
     }
 
     fn annotate_circuit(&self, region: &mut Region<F>) {
@@ -337,9 +356,17 @@ pub struct ChiquitoHalo2Circuit<F: Field + From<u64>> {
     witness: Option<Assignments<F>>,
 }
 
-impl<F: Field + From<u64>> ChiquitoHalo2Circuit<F> {
+impl<F: Field + From<u64> + Hash> ChiquitoHalo2Circuit<F> {
     pub fn new(compiled: ChiquitoHalo2<F>, witness: Option<Assignments<F>>) -> Self {
         Self { compiled, witness }
+    }
+
+    pub fn instance(&self) -> Vec<Vec<F>> {
+        if let Some(witness) = &self.witness {
+            vec![self.compiled.instance(witness)]
+        } else {
+            Vec::new()
+        }
     }
 }
 
--- a/src/plonkish/backend/halo2.rs
+++ b/src/plonkish/backend/halo2.rs
@@ -415,11 +415,33 @@ pub struct ChiquitoHalo2SuperCircuit<F: Field + From<u64>> {
     witness: SuperAssignments<F>,
 }
 
-impl<F: Field + From<u64>> ChiquitoHalo2SuperCircuit<F> {
+impl<F: Field + From<u64> + Hash> ChiquitoHalo2SuperCircuit<F> {
     pub fn new(sub_circuits: Vec<ChiquitoHalo2<F>>, witness: SuperAssignments<F>) -> Self {
         Self {
             sub_circuits,
             witness,
+        }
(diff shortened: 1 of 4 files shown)


leolara:
@qwang98 I don't understand, the num_steps is set in the python frontend, and the padding is done in the python frontend. So I don't understand what you mean about getting num_steps from AST. The AST is in the python frontend where the padding is done.

qwang98:
Sorry for the confusion. My description was incorrect and doesn't match the changes. The changes are:
1. Store AST in the CIRCUIT_MAP, so that the MockProver can obtain whether there's exposed columns in the AST, and fill in the third `instance` parameter of `MockProver::run` accordingly, i.e. empty Vec if there's no instance or non-empty Vec if there is.
3. Added a debug function that prints witness after converting it to Plaf. This was quite helpful for identifying a couple bugs that I fixed. I exposed it in PyO3 but not in Python DSL, as it's not some function the user will use.
4. Added empty trace function for AST Circuit deserializer, because otherwise the compiler won't generate witness if trace defaults to None.

@leolara Ready for review again :)

leolara:
> 1. Store AST in the CIRCUIT_MAP, so that the MockProver can obtain whether there's exposed columns in the AST, and fill in the third `instance` parameter of `MockProver::run` accordingly, i.e. empty Vec if there's no instance or non-empty Vec if there is.

So the solution for this is explained here: https://github.com/privacy-scaling-explorations/chiquito/issues/83 and also the exposed column is in the plonkish IR and in the ChiquitoHalo2 so it wouldn't be necessary to keep the AST for that.

leolara:
> 2. Added a debug function that prints witness after converting it to Plaf. This was quite helpful for identifying a couple bugs that I fixed. I exposed it in PyO3 but not in Python DSL, as it's not some function the user will use.

Why is there plaf stufff in the python frontend? I think for now we should not implement plaf backend for the python frontend

leolara:
> 2. Added a debug function that prints witness after converting it to Plaf. This was quite helpful for identifying a couple bugs that I fixed. I exposed it in PyO3 but not in Python DSL, as it's not some function the user will use.

I think this is better solved by implementing Debug and/or Display traits on certain types.

qwang98:
> > 1. Store AST in the CIRCUIT_MAP, so that the MockProver can obtain whether there's exposed columns in the AST, and fill in the third `instance` parameter of `MockProver::run` accordingly, i.e. empty Vec if there's no instance or non-empty Vec if there is.
> 
> So the solution for this is explained here: #83 and also the exposed column is in the plonkish IR and in the ChiquitoHalo2 so it wouldn't be necessary to keep the AST for that.

Cool. Working on this now :)

qwang98:
> > 2. Added a debug function that prints witness after converting it to Plaf. This was quite helpful for identifying a couple bugs that I fixed. I exposed it in PyO3 but not in Python DSL, as it's not some function the user will use.
> 
> Why is there plaf stufff in the python frontend? I think for now we should not implement plaf backend for the python frontend

It's not implementing Plaf backend. It's really just printing witness assignment with the help of Plaf.

qwang98:
> > 2. Added a debug function that prints witness after converting it to Plaf. This was quite helpful for identifying a couple bugs that I fixed. I exposed it in PyO3 but not in Python DSL, as it's not some function the user will use.
> 
> I think this is better solved by implementing Debug and/or Display traits on certain types.

I see your point now. I'll think about it and either implement this or remove the print_plaf_witness function entirely. :)

leolara:
The optimal solution would be to implement unit tests

leolara:
> > > 2. Added a debug function that prints witness after converting it to Plaf. This was quite helpful for identifying a couple bugs that I fixed. I exposed it in PyO3 but not in Python DSL, as it's not some function the user will use.
> > 
> > 
> > Why is there plaf stufff in the python frontend? I think for now we should not implement plaf backend for the python frontend
> 
> It's not implementing Plaf backend. It's really just printing witness assignment with the help of Plaf.

We should implement the Debug and Display for TraceWitness struct, Assigments, etc... much easier that way. But can be done in another PR

leolara:
@qwang98 I think we can make it for the super circuit, if we follow the same order as the meta.instance_columns are called.

src/plonkish/backend/halo2.rs
```diff
@@ -388,13 +415,35 @@ pub struct ChiquitoHalo2SuperCircuit<F: Field + From<u64>> {
...
     pub fn new(sub_circuits: Vec<ChiquitoHalo2<F>>, witness: SuperAssignments<F>) -> Self {
         Self {
             sub_circuits,
             witness,
         }
     }
+
+    pub fn instance(&self) -> Vec<Vec<F>> {
+        if !self.witness.is_empty() {
+            let mut result = Vec::new();
```
leolara, line 428: if you put let mut result outside if, you don't need the else. Just write result in the last line

src/plonkish/backend/halo2.rs
```diff
@@ -388,13 +415,35 @@ pub struct ChiquitoHalo2SuperCircuit<F: Field + From<u64>> {
...
+
+            for (uuid, assignment) in &self.witness {
```
leolara, line 430: @qwang98 I don't think you should follow the order of the witness, but the order of sub_circuit, or whatever order is used to call meta.instance_column()

qwang98: Agreed. sub_circuit is the order meta.instance_column is called.

src/plonkish/backend/halo2.rs
qwang98: Indeed.

qwang98:
> @qwang98 I think we can make it for the super circuit, if we follow the same order as the meta.instance_columns are called.

Done

qwang98:
> > > > 2. Added a debug function that prints witness after converting it to Plaf. This was quite helpful for identifying a couple bugs that I fixed. I exposed it in PyO3 but not in Python DSL, as it's not some function the user will use.
> > > 
> > > 
> > > Why is there plaf stufff in the python frontend? I think for now we should not implement plaf backend for the python frontend
> > 
> > 
> > It's not implementing Plaf backend. It's really just printing witness assignment with the help of Plaf.
> 
> We should implement the Debug and Display for TraceWitness struct, Assigments, etc... much easier that way. But can be done in another PR

Agree on this.

leolara (approved):

PR #89, opened 2023-08-22
Develop keccak

10to4:


Cargo.toml
```diff
@@ -20,6 +20,7 @@ num-bigint = { version = "0.4", features = ["rand"] }
 uuid = { version = "1.4.0", features = ["v1", "rng"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+ff = "0.13.0"
```
leolara, line 23: @10to4 why is this package required, it seems it is not used

10to4: Yes, I think this should be legacy useless code.

leolara (approved):

leolara:
@10to4 I have approved it, perhaps you can fix those clippy errors and merge it :-)

PR #107, opened 2023-08-23
Steve/update pychiquito

qwang98:
Now ready for review. I apologize for a complicated PR as they should be broken up further in retrospect. Changes include:

**Tutorials**
- All Fibo tutorials now work (through chapter 5)
- Added padding related features in Python frontend to support new tutorial chapters
- Published latest Chiquito Python package 0.2.1 to PyPI
- Updated README and associated links (including a main README and a Python frontend README)

**Changes Related to Fixed Gen**
- To parse Python generated fixed assignments directly to Rust, I moved fixed assignments generation from compiler to dsl/ast
- As a result, Rust ast no longer stores `fixed_gen` lambda, but is replaced with `fixed_assignments`
- Added associated functions in Python

**New (Incomplete) Python Example: Mimc7**
- Mimc7 example works for the mimc7_constants circuit
- To support this example, Lookup is now enabled in Python frontend and can be parsed to Rust ast
- The main circuit is still under development and will be pushed in another PR for Python super circuit


qwang98:
> I am sorry Steve, this needs to be broken down. Please, think about the order in which these things depend on each other and send the PRs in that order.
> 
> Also, we need to decide about the versioning. I don't think next version should be 0.2.1. I think we should keep the vYYYY_MM_DD if that is possible in pypi.
> 
> Then when it is more stable we can for another versioning

I tried a few times and versioning in the format of `vYYYY_MM_DD` or `YYYY_MM_DD` won't work. We run into build errors from the CI.

leolara:
I think v0.YYYY_MM_DD would be better so in the future we can switch.

Perhaps we can do v0.1.YYYYMMDD for now and use the same for the branches. Are you aware of the Tran PR on this?

qwang98:
> I think v0.YYYY_MM_DD would be better so in the future we can switch.
> 
> Perhaps we can do v0.1.YYYYMMDD for now and use the same for the branches. Are you aware of the Tran PR on this?

I think we can do 0.2.20230828 for now. Unfortunately I already published 0.2.0 and 0.2.1

leolara:
We can delete the published versions I think. Also I think we need to add two additional 00 at the end in case there is more than one version in the same day

qwang98:
> We can delete the published versions I think. Also I think we need to add two additional 00 at the end in case there is more than one version in the same day

kk this works.

qwang98:
Tutorial PR ready for review at #115 

I will break up the rest later.

PR #108, opened 2023-08-23
Github workflow to publish pypi python package

### LINKED ISSUE #91: Give instructions or add to CI for updating pip package

trangnv:
Resolve #91 
- We need to have PYPI_API_TOKEN as repo secret
- workflow will run with tag "v*" e.g. `git tag v1.0.0`

### DIFF
--- a/.github/workflows/pypi_publish.yml
+++ b/.github/workflows/pypi_publish.yml
@@ -1,0 +1,126 @@
+name: Pypi Publish
+
+on:
+  push:
+    branches:
+      - main
+      - master
+    tags:
+      - "v*"
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  linux:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target: [x86_64, x86, aarch64, armv7, s390x, ppc64le]
+    steps:
+      - name: Checkout repository and submodules
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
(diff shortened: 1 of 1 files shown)


leolara:
@trangnv there is a CI task failing, is it because PYPI_API_TOKEN is not set?

leolara:
> workflow will run with tag "v*" e.g. git tag v1.0.0

does this mean that the package is only published when a tag is added?

trangnv:
> @trangnv there is a CI task failing, is it because PYPI_API_TOKEN is not set?

That's right, I dont have the right to set the repo secret.


> > workflow will run with tag "v*" e.g. git tag v1.0.0
> 
> does this mean that the package is only published when a tag is added?

Yes, we might not want to build and publish python package everytime we merge. But let me check again with the forked one if that works as intended.

leolara:
> > @trangnv there is a CI task failing, is it because PYPI_API_TOKEN is not set?
> 
> 
> 
> That's right, I dont have the right to set the repo secret.
> 
> 
> 
> 
> 
> > > workflow will run with tag "v*" e.g. git tag v1.0.0
> 
> > 
> 
> > does this mean that the package is only published when a tag is added?
> 
> 
> 
> Yes, we might not want to build and publish python package everytime we merge. But let me check again with the forked one if that works as intended.

Thanks I think for now we would like to use as versions the following format: vYYYY_MM_DD

leolara:
Actually we decided on v0.1.YYYYMMDDrr where rr is 00 or 01 if we need to release more than once a day

leolara (approved):

leolara:
@trangnv you can merge now, thanks

leolara:
I am going to merge this because another PR needs this to be merged before continuing

PR #111, opened 2023-08-28
feat: add pretty debug information for common items

### LINKED ISSUE #85: Debug and/or Display traits for common items such as TraceWitness and Assignment
This is good for debugging. Add unit tests when appropriate.

dyxushuai:
Resolve #85

- [x] TraceWitness 
- [x] Assignments


dyxushuai:
@leolara @ChihChengLiang I am keen to know if we only need print information with well indentation. Just like what format's `{:#}` flag's do 
https://doc.rust-lang.org/std/fmt/#sign0

leolara:
I think we should make display look more like text, so for example

```
00(9): a = 1, b =2
01(10): a = 1, b =2
```

offset(step_uuid): assignations

where offset and step_uuid are hexadecimal.

dyxushuai:
> I think we should make display look more like text, so for example
> 
> ```
> 00(9): a = 1, b =2
> 01(10): a = 1, b =2
> ```
> 
> offset(step_uuid): assignations
> 
> where offset and step_uuid are hexadecimal.

Ok, Make sense

src/wit_gen.rs
```diff
@@ -39,6 +39,44 @@ pub struct TraceWitness<F> {
...
+                |mut acc, (queriable, value)| {
+                    acc.push_str(&format!("{:?} = {:?}, ", queriable, value));
+                    acc
+                },
+            );
+            // get the decimal width based on the assignments size, and extra one leading zero
+            // len in 0..=9 -> width = 2
+            // len in 10..=99-> width = 3
+            // len in 100..=999 -> width = 4
+            let decimal_width = step_instance
```
dyxushuai, line 60: The output of the length of offset's padding is dynamic, does this make sense?

dyxushuai: Or, other fields like `step_uuid` needs  to be dynamic width.

dyxushuai:
After the format of `TraceWitness` is OK, I will continue work

qwang98:
Just bumping this up. @leolara

leolara:
@dyxushuai continue, I think it is ok, perhaps we don't need so detailed commenting, just showing the expected output.

leolara:
@dyxushuai can you continue with this or is it finished?

dyxushuai:
@leolara Pls take another look :)

leolara (approved):

PR #115, opened 2023-08-28
Finalize fibo tutorials

qwang98:
Ready for review.
- All Fibo tutorials now work (through chapter 5)
- Added padding related features in Python frontend to support new tutorial chapters
- Published latest Chiquito Python package `0.1.2023082800` to PyPI


leolara:
@qwang98 I think this conflicts with @trangnv previous work on the CI, can you remove that so the CI comes from @trangnv PR?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "chiquito"
-version = "0.1.0"
+version = "0.1.2023082800"
```
leolara, line 3: @qwang98 @trangnv do we need to update this for each release, or can the release come from the git tag?

requirements.txt
```diff
@@ -0,0 +1,6 @@
+py-ecc==6.0.0
```
leolara, line 1: @qwang98 @trangnv should this be in the root folder or in src/frontend/python?

src/frontend/python/chiquito/dsl.py
```diff
@@ -84,9 +85,29 @@ def pragma_disable_q_enable(self: Circuit) -> None:
...
+        return len(self.witness.step_instances) < self.ast.num_steps
+
+    def padding(self: Circuit, step_type: StepType, args: Any):
+        assert self.mode == CircuitMode.Trace
+        while self.needs_padding():
+            self.add(step_type, args)
+
+    # called under fixed_gen()
+    def assign(self: Circuit, offset: int, lhs: Queriable, rhs: F):
+        assert self.mode == CircuitMode.FixedGen
```
leolara, line 104: @qwang98 we don't have this mode in this PR

leolara:
@qwang98 why do we need to publish pypi here? the tutorial is not installing `pip install chiquito` right?

qwang98:
> @qwang98 I think this conflicts with @trangnv previous work on the CI, can you remove that so the CI comes from @trangnv PR?

Sure :)

requirements.txt
qwang98: Technically the entire requirements.txt is for Python. I think putting them at the root folder is easier for users to call `pip install -r requirements.txt`, but let me know if you think otherwise.

src/frontend/python/chiquito/dsl.py
qwang98: Yes. Will put in another PR.

qwang98: Removed.

requirements.txt
leolara: Ok

Cargo.toml
leolara: Cargo.toml triggers the pypi publication?

So to have a tag is a good practice, so we should have that. If it is still necessary to update this, perhaps we can have a script that creates the tag and also updates this with `sed` or from a template or something like that.

But would like to get explained how the CI pypi publication is triggered and create a markdown that explains how to create a release.

leolara:
> > @qwang98 why do we need to publish pypi here? the tutorial is not installing `pip install chiquito` right?
> 
> It's installed with `pip install -r requirements.txt`. Chiquito is listed as one of the dependencies. I made updates such as padding that are used in chapter 5 and that's why we need to update the PyPI package.

But this is not done to make the tutorial work right? Or this have to be done before creating the kernel?

leolara (approved):

qwang98:
> > > @qwang98 why do we need to publish pypi here? the tutorial is not installing `pip install chiquito` right?
> > 
> > 
> > It's installed with `pip install -r requirements.txt`. Chiquito is listed as one of the dependencies. I made updates such as padding that are used in chapter 5 and that's why we need to update the PyPI package.
> 
> But this is not done to make the tutorial work right? Or this have to be done before creating the kernel?

Yes they are needed for creating the kernel, which copies over the local Python environment.

PR #117, opened 2023-08-29
Python: fixed gen, lookup, and mimc7

qwang98:
Ready for review @leolara.

**Changes Related to Fixed Gen**
- To parse Python generated fixed assignments directly to Rust, I moved fixed assignments generation from compiler to dsl/ast
- As a result, Rust ast no longer stores fixed_gen lambda, but is replaced with fixed_assignments
- Added associated functions in Python
- New (Incomplete) Python Example: Mimc7

**Mimc7 example works for the mimc7_constants circuit**
- To support this example, Lookup is now enabled in Python frontend and can be parsed to Rust ast
- The main circuit is still under development and will be pushed in another PR for Python super circuit
- Won't close the Mimc7 PR till we have supercircuit done


leolara:
Please constants at the bottom if python allows it, even better if the constants can be in another file.

qwang98:
> Please constants at the bottom if python allows it, even better if the constants can be in another file.

Putting at bottom won't work, event with `__future__` imported. I'll try another file.

src/frontend/python/chiquito/dsl.py
```diff
@@ -1,31 +1,50 @@
...
         self.rust_ast_id = 0
         self.mode = CircuitMode.SETUP
         self.setup()
+        if hasattr(self, "fixed_gen") and callable(self.fixed_gen):
+            self.mode = CircuitMode.FixedGen
+            if self.ast.num_steps == 0:
+                raise ValueError(
+                    "Must set num_steps by calling pragma_num_steps() in setup before calling fixed_gen()."
+                )
+            self.fixed_gen_context = FixedGenContext.new(self.ast.num_steps)
```
leolara, line 44: @qwang98 I don't think we need the class FixedGenContext anymore, we could keep the assignations array in this class

qwang98: Agreed. I kept fixed_assignments in AST as we did in Rust, and added AST function that modifies it. DSL calls this AST function.

qwang98:
Addressed Leo's comments. Now still runs successfully on mimc7 example. Ready for review again. @leolara

leolara:
sorry, I had comments but I didn't submit them

src/ast/mod.rs
```diff
@@ -167,21 +179,33 @@ impl<F, TraceArgs> Circuit<F, TraceArgs> {
...
-    pub fn set_fixed_gen<D>(&mut self, def: D)
+    pub fn get_step_type(&self, uuid: UUID) -> Rc<StepType<F>> {
+        let step_rc = self.step_types.get(&uuid).expect("step type not found");
+
+        Rc::clone(step_rc)
+    }
+}
+
+impl<F: Field + Hash, TraceArgs> Circuit<F, TraceArgs> {
+    pub fn set_fixed_assignments<D>(&mut self, def: D)
```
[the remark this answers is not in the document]
qwang98: Resolved!

src/frontend/python/chiquito/cb.py
```diff
@@ -222,3 +227,93 @@ def to_constraint(v: ToConstraint) -> Constraint:
...
+        return self
+
+    def when(self: LookupTableBuilder, enable: ToConstraint) -> LookupTableBuilder:
+        if self.enable is not None:
+            raise ValueError("LookupTableBuilder: when() can only be called once.")
+        self.enable = to_constraint(enable)
+        return self
+
+    def build(self: LookupTableBuilder, step_type: StepType) -> Lookup:
+        table = step_type.tables.get(self.id)
```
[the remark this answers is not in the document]
qwang98: In Rust Chiquito, table is initially built using `new_table` from DSL and stored in `CircuitContext`. It's then cloned to `StepTypeContext` and `StepTypeSetupContext`. Because `build` is called by `add_lookup`, a `StepTypeSetupContext` function, `build` conveniently takes a `ctx: StepTypeSetupContext` parameter. Therefore, table doesn't "have to be" stored in the step, but it's just the most convenient for `build` to have the `step_type` parameter, as `add_lookup` calls `build` at the step type level.

In Python, `add_lookup` also calls `build` at the step type level. The tables are initially created and stored at the circuit level, but step type and super circuit can also have references to the tables.

qwang98:
Ready for review again. @leolara

src/ast/mod.rs
```diff
@@ -167,23 +179,24 @@ impl<F, TraceArgs> Circuit<F, TraceArgs> {
...
-
     pub fn get_step_type(&self, uuid: UUID) -> Rc<StepType<F>> {
         let step_rc = self.step_types.get(&uuid).expect("step type not found");
 
         Rc::clone(step_rc)
     }
 }
 
+impl<F: Field + Hash, TraceArgs> Circuit<F, TraceArgs> {
+    pub fn set_fixed_assignments(&mut self, assignments: Option<FixedAssignment<F>>) {
```
leolara, line 190: @qwang98 why is the parameter an option? Would you call it with None?

src/frontend/python/chiquito/cb.py
```diff
@@ -222,3 +227,93 @@ def to_constraint(v: ToConstraint) -> Constraint:
         raise TypeError(
             f"Type `{type(v)}` is not ToConstraint (one of Constraint, Expr, int, or F)."
         )
+
+
+@dataclass
+class LookupTableRegistry:
```
leolara, line 233: @qwang98 I don't think we need a LookupTableRegistry, this was necessary in rust to have the `Copy` constraint, but in python the table is a reference.

src/frontend/python/chiquito/cb.py
```diff
@@ -222,3 +227,93 @@ def to_constraint(v: ToConstraint) -> Constraint:
...
+
+    def add(self: LookupTableRegistry, table: LookupTable):
+        self.tables[table.uuid] = table
+
+    def get(self: LookupTableRegistry, uuid: int) -> LookupTable:
+        return self.tables[uuid]
+
+
+@dataclass
+class LookupTable:
```
leolara, line 244: @qwang98 when a table is created, it should be close to add more expresions to dest. So perhaps there is a method create() that returns self and sets a flag that prevents from adding more things to dest. Also, the table cannot be used if this flag has not been activated by call to create()

src/ast/mod.rs
qwang98: Fixed!

src/frontend/python/chiquito/cb.py
qwang98: Removed and fixed!

src/frontend/python/chiquito/cb.py
qwang98: I believe this `create()` function you mentioned is actually `new_table()` in `dsl.py`. Rust Chiquito has the same naming.

I modified `LookupTable` such that now we have a new `read_only` flag, that defaults to `False` when `LookupTable` is just created. User can then call `add()` to add more destination columns to the `LookupTable`, which requires that `read_only == False`. 

`LookupTable` will then need to be added by `new_table()` in `dsl.py` to the dictionary of UUID to `LookupTable`. `new_table()` also changes the `read_only` flag to `True`, so that no more calls of `add()` to modify the destination column is possible after `new_table()` is called.

Finally, any use of the table will be through `apply()` and `when()`, both of which requires that `read_only == True`.

This should accomplish all the features you suggested above.

What's more? `create_table()` can only be called during `setup()` and there's a flag assertion for that as well.

qwang98:
Everything fixed and ready for review again. @leolara

qwang98:
All contents will be PR'ed through #125 instead. Closing this.

PR #120, opened 2023-09-06
New polynomial expressions system

leolara:
It allows:
 + One polynomial expression implementation for frontend and IR
 + In the future compilation of arbitrary boolean expressions


qwang98:
Reviewing this now.

src/field.rs
```diff
@@ -0,0 +1,46 @@
...
+    + AddAssign
+    + SubAssign
+    + MulAssign
+    + for<'a> AddAssign<&'a Self>
+    + for<'a> SubAssign<&'a Self>
+    + for<'a> MulAssign<&'a Self>
+    + Send
+    + Sync
+    + Debug
+    + 'static
```
qwang98, line 34: Compared to `halo2proofs::arithmetic::Field`:

Removed:
    + ConditionallySelectable
    + ConstantTimeEq

Any specific reasons?

src/frontend/dsl/sc.rs
```diff
@@ -1,6 +1,6 @@
 use std::{hash::Hash, rc::Rc};
 
-use halo2_proofs::arithmetic::Field;
+use crate::field::Field;
```
qwang98, line 3: I did a search over the whole project. Looks like we still have halo2_proofs::arithmetic::Field in halo2.rs and fibonacci.rs.

I think we should also replace them?

src/plonkish/backend/halo2.rs
```diff
@@ -11,17 +11,27 @@ use halo2_proofs::{
...
 
+impl<T: Field + From<u64>> ChiquitoField for T {
+    const ZERO: Self = <Self as Field>::ZERO;
+    const ONE: Self = <Self as Field>::ONE;
+
+    fn mi(&self) -> Self {
+        self.invert().unwrap_or(Self::ZERO)
+    }
+}
+
```
qwang98, line 34: I don't see where we are using `T`.

Besides, looks like the only difference from ff::Field is that we return ZERO instead of error if the element inverted is zero.

What's the purpose of this again? To avoid using the is zero gadget?

src/plonkish/compiler/mod.rs
```diff
@@ -525,13 +512,10 @@ fn add_q_last_to_constraint<F: Field>(
     constraint: PolyExpr<F>,
 ) -> PolyExpr<F> {
     let q_last_column = unit.last_step.clone().expect("last column not found").1;
-    let q_last = PolyExpr::<F>::Query(q_last_column, 0, "q_last".to_owned());
-    let not_q_last_expr = PolyExpr::Sum(vec![
-        PolyExpr::Const(F::ONE),
-        PolyExpr::Neg(Box::new(q_last)),
-    ]);
+    let q_last = q_last_column.query(0, "q_last".to_owned());
+    let not_q_last_expr = PolyExpr::Const(F::ONE) + (-q_last);
```
qwang98, line 517: Saw this construction a few times, but what's the difference between -q-last and + (-q_last)?

src/stdlib.rs
```diff
@@ -46,6 +47,6 @@ impl<F: Field + From<u64>> IsZero<F> {
 
 impl<F: Field + Eq + Hash> IsZero<F> {
     pub fn wg<WGC>(&self, ctx: &mut StepInstance<F>, value: F) {
```
qwang98, line 49: Probably missed this from some other PR but why do we need a <WGC> generic parameter here that's unused?

src/stdlib.rs
```diff
@@ -46,6 +47,6 @@ impl<F: Field + From<u64>> IsZero<F> {
...
-        ctx.assign(self.value_inv, value.invert().unwrap_or(F::ZERO));
+        ctx.assign(self.value_inv, value.mi());
```
qwang98, line 50: The only `mi` implementation is for `T` in `halo2.rs`, but I don't see T used anywhere. Am I missing something?

qwang98:
Overall, I do see the benefits in terms of syntax simplification of unifying PolyExpr and ASTExpr as Expr<F, Queriable> and Expr<F, Queriable<F>>, respectively, where Queriable has different definitions under AST and IR.

However, I'm a bit unfamiliar with the motivation behind creating our own Field with alternative implementation of multiplicative inverse. How is this related to future compilations of arbitrary boolean expressions?

qwang98 (approved):
Mostly soft comments except my comments about `T` and `Field`. Should be good to go once resolved. :)

src/field.rs
leolara: We don't use these traits in our code.

src/frontend/dsl/sc.rs
leolara: I changed the fibo example to use our Field, but it does not matter. There is no relation with the inverse thing.

halo2 needs to use its trait

src/plonkish/backend/halo2.rs
leolara: the purpose is that the core chiquito should not depend on halo2 traits as it should work with other backends.

This is a blanket implementation. It means that the trait is implemented for all Ts such that T: Field + From<u64>.

https://users.rust-lang.org/t/what-are-blanket-implementations/49904

src/plonkish/compiler/mod.rs
leolara: None, just for clarity and avoiding problems with the operator overloading

src/stdlib.rs
leolara: it seems this is unrelated, please create an issue

src/stdlib.rs
leolara: No, the reason is that our core code should not depend on halo2, so it is compatible with other backends.

PR #125, opened 2023-09-10
Python Supercircuit

qwang98:
Will be PR'ed to main.

@leolara Ready for review and the MiMC7 example is fully functional.

# My technical approach
- Pass over JSON of sub-circuit AST to Rust. Obtain uuid.
- Pass over JSON of sub-circuit TraceWitness to Rust using the same uuid.
- `halo2_mock_prover` function takes vector of uuids and compiles super circuit before passing it to the mock prover.


src/frontend/python/chiquito/chiquito_ast.py
```diff
@@ -418,19 +427,19 @@ def add(
...
 
     def enable(self: Lookup, enable_annotation: str, enable_expr: Expr):
-        enable = ASTConstraint(enable_annotation, enable_expr)
-        if self.enable is None:
+        enabler = ASTConstraint(enable_annotation, enable_expr)
+        if self.enabler is None:
             for constraint, _ in self.exprs:
-                constraint = self.multiply_constraints(enable, constraint)
-            self.enable = enable
+                constraint = self.multiply_constraints(enabler, constraint)
```
qwang98, line 441: Called `enabler` here so that it doesn't clash with the function name.

src/frontend/python/chiquito/dsl.py
```diff
@@ -99,7 +168,11 @@ def pragma_disable_q_enable(self: Circuit) -> None:
 
     def new_table(self: Circuit, table: LookupTable) -> LookupTable:
         assert self.mode == CircuitMode.SETUP
-        self.tables.add(table)
+        if self.super_circuit is None:
+            raise SyntaxError(
+                "Circuit: new_table() is only available for Circuit with initiated super_circuit field."
+            )
+        self.super_circuit.tables[table.uuid] = table
```
qwang98, line 175: Note that here I opted to store the tables under supercircuit and then circuit/steptype both have reference to the supercircuit, so they can reference the tables as well.

qwang98:
@leolara All bugs fixed and fully functional. :)

leolara:
1. I don't think we need to keep the tables in each circuit like in the rust front-end. It is a technical thing for rust, and we can send that as an empty array. The key is that the lookup tables are already built, into actual lookups when used in the python side, so we actually send step_type.lookups, and that is all that the rust side needs, the LookupTableRegistry is not needed anymore.

leolara:
2. I think should should create a new py03 endpoint, chiquito_ast_sc_to_halo2, that receives an array of AST.

Then with the array of AST you can just create a `SuperCircuitContext`, add all sub circuits with `sub_circuit` and then call `compile`. That should work.

leolara:
3. For the mapping, you should serialise a SuperAssginments. You are doing something complicated with the uuid, just make sure that the AST uuid generated in the python side can be used to index the SuperAssigments.

qwang98:
> 1. I don't think we need to keep the tables in each circuit like in the rust front-end. It is a technical thing for rust, and we can send that as an empty array. The key is that the lookup tables are already built, into actual lookups when used in the python side, so we actually send step_type.lookups, and that is all that the rust side needs, the LookupTableRegistry is not needed anymore.

That's actually exactly what I did. I'd note that I didn't serialize `SuperCircuitContext` or `LookupTable`, but I did serialize `Lookup` in `StepType`. I also removed `LookupTableRegistry` from Python and replaced it with a Dict[UUID, LookupTable] under `SuperCircuit`. Again, this is only used on the Python end to create `Lookup` but never sent to the Rust end. The Rust end only ever receives `Lookup` that are already built. :)

qwang98:
> 2. I think should should create a new py03 endpoint, chiquito_ast_sc_to_halo2, that receives an array of AST.
> 
> Then with the array of AST you can just create a `SuperCircuitContext`, add all sub circuits with `sub_circuit` and then call `compile`. That should work.

This is actually what I did as well! There's a `chiquito_super_circuit_halo2_mock_prover(rust_ids: Vec<UUID>)` function that takes an array of `rust_ids`, each of which maps to an AST (together with the corresponding `TraceWitness` and `AssignmentGenerator`). Then, the function creates a `SuperCircuitContext` and adds all sub-circuits with `sub_circuit_with_ast`, which takes an AST instead of a closure. After that, the function calls compile on the `SuperCircuitContext`.

Note that the ASTs are already passed to Rust with `rust_ids` generated for them when the user calls `sub_circuit` from Python.

I'm marking the `chiquito_super_circuit_halo2_mock_prover` Rust function and the `sub_circuit` Python function each as a comment here for your convenience.

src/frontend/python/mod.rs
```diff
@@ -46,25 +58,101 @@ pub fn chiquito_ast_to_halo2(ast_json: &str) -> UUID {
...
+    let result = prover.verify_par();
+
+    println!("result = {:#?}", result);
+
+    if let Err(failures) = &result {
+        for failure in failures.iter() {
+            println!("{}", failure);
+        }
+    }
+}
```
qwang98, line 139: `chiquito_super_circuit_halo2_mock_prover`

src/frontend/python/chiquito/dsl.py
```diff
@@ -19,6 +21,70 @@
...
+    def sub_circuit(self: SuperCircuit, sub_circuit: Circuit) -> Circuit:
+        assert self.mode == SuperCircuitMode.SETUP
+        if sub_circuit.rust_id != 0:
+            raise ValueError(
+                "SuperCircuit: sub_circuit() cannot be called twice on the same circuit."
+            )
+        ast_json: str = sub_circuit.get_ast_json()
+        sub_circuit.rust_id: int = rust_chiquito.ast_to_halo2(ast_json)
+        self.ast.sub_circuits[sub_circuit.rust_id] = sub_circuit.ast
+        return sub_circuit
```
qwang98, line 49: `sub_circuit` Python function.

qwang98:
> 3. For the mapping, you should serialise a SuperAssginments. You are doing something complicated with the uuid, just make sure that the AST uuid generated in the python side can be used to index the SuperAssigments.

I'd comment that historically we only serialize a `TraceWitness`, pass it over to Rust, and then process it with an `AssignmentGenerator` to obtain `Assignments`. We don't serialize `Assignments` directly, as it requires Python versions of Rust functions like `assign_step`, `find_placement`, and `get_selector_assignment` from `assignment.rs`, which are better left to Rust alone. Therefore, I wonder by saying serialize `SuperAssignments`, which is a mapping of UUID to `Assignments`, you actually meant to serialize a mapping UUID to `TraceWitness`?

That's actually what I did, to serialize a mapping of `TraceWitness` and then associate it with the corresponding AST under the same UUID. To walk you through the steps, here's the MiMC7 example:

```
mimc7 = Mimc7SuperCircuit()
mimc7_witnesses = mimc7.gen_witness((F(1), F(2)))
mimc7.halo2_mock_prover(mimc7_witnesses)
```

We first call `gen_witness` on the `SuperCircuit` `mimc7`, which returns a `Dict[UUID, TraceWitness]`. This `TraceWitness` dictionary is then passed to `halo2_mock_prover`, which serializes each `TraceWitness`, send it to Rust, and store it under the same UUID as its corresponding AST.

Then in Rust, the `chiquito_super_circuit_halo2_mock_prover` function converts each `TraceWitness` to an `Assignments`, using the `AssignmentGenerator` with the same UUID. These conversions are all done under a `MappingContext`, which eventually spits out a `SuperAssignments`.

Just to clarify, I created a `UUID` -> `CircuitMapStore` master mapping in Rust for the Python frontend, where `CircuitMapStore` is defined as:
```
type CircuitMapStore = (
    Circuit<Fr, ()>,
    ChiquitoHalo2<Fr>,
    Option<AssignmentGenerator<Fr, ()>>,
    Option<TraceWitness<Fr>>,
);
```

This is how all corresponding items are grouped under the same `UUID`.

Again, for your convenience, I'm marking the `gen_witness` Python function and `halo2_mock_prover` Python function each as a comment here.

src/frontend/python/chiquito/dsl.py
```diff
@@ -19,6 +21,70 @@
...
+    # generates TraceWitness mapping
+    def gen_witness(self: SuperCircuit, args: Any) -> Dict[int, TraceWitness]:
+        self.mode = SuperCircuitMode.Mapping
+        self.mapping(args)
+        self.mode = SuperCircuitMode.NoMode
+        witnesses: Dict[int, TraceWitness] = self.ast.witnesses
+        del (
+            self.ast.witnesses
+        )  # so that we can generate different witness mapping in the next gen_witness() call
+        return witnesses
```
qwang98, line 73: `gen_witness` Python function for `SuperCircuit`

src/frontend/python/chiquito/dsl.py
```diff
@@ -19,6 +21,70 @@
...
+        for rust_id, witness in witnesses.items():
+            witness_json: str = witness.get_witness_json()
+            if rust_id not in self.ast.sub_circuits:
+                raise ValueError(
+                    f"SuperCircuit.halo2_mock_prover(): TraceWitness with rust_id {rust_id} not found in sub_circuits."
+                )
+            rust_chiquito.add_witness_to_rust_id(witness_json, rust_id)
+        rust_chiquito.super_circuit_halo2_mock_prover(
+            list(self.ast.sub_circuits.keys())
+        )
```
qwang98, line 85: `halo2_mock_prover` Python function for `SuperCircuit`

qwang98:
Ready for review again. Mostly responding to comments. I also merged latest changes. @leolara

examples/mimc7.py
```diff
@@ -0,0 +1,169 @@
...
+            )
+        )
+
+        self.transition(eq(self.y, self.circuit.x.next()))
+        self.transition(eq(self.circuit.k, self.circuit.k.next()))
+        self.transition(eq(self.circuit.row, 0))
+        self.transition(eq(self.circuit.row + 1, self.circuit.row.next()))
+
+        self.add_lookup(
+            self.circuit.imports.apply(self.circuit.row).apply(self.circuit.c)
```
leolara, line 89: @qwang98 I think it makes sense to call it self.circuit.constants_table

examples/mimc7.py
```diff
@@ -0,0 +1,169 @@
...
+            self.add(self.mimc7_step, x_value, k_value, c_value, row_value)
+
+        row_value += F(1)
+        x_value += F(k_value + c_value)
+        x_value = F(x_value**7)
+
+        self.add(self.mimc7_last_step, x_value, k_value, c_value, row_value)
+
+
+class Mimc7FirstStep(StepType):
```
leolara, line 64: @qwang98 I think it is more clear to put the steps before the circuit

src/frontend/dsl/mod.rs
```diff
@@ -172,6 +162,27 @@ impl<F, TraceArgs> CircuitContext<F, TraceArgs> {
     }
 }
 
+impl<F: Field + Hash, TraceArgs> CircuitContext<F, TraceArgs> {
+    /// Sets the fixed generation function for the circuit. The fixed generation function is
```
leolara, line 166: @qwang98 this does not "set", but "executes" and sets the assignations

src/frontend/python/mod.rs
```diff
@@ -27,13 +30,22 @@ use halo2_proofs::{dev::MockProver, halo2curves::bn256::Fr};
 use serde::de::{self, Deserialize, Deserializer, IgnoredAny, MapAccess, Visitor};
 use std::{cell::RefCell, collections::HashMap, fmt, rc::Rc};
 
-type CircuitMapStore = (ChiquitoHalo2<Fr>, Option<AssignmentGenerator<Fr, ()>>);
+type CircuitMapStore = (
+    Circuit<Fr, ()>,
+    ChiquitoHalo2<Fr>,
+    Option<AssignmentGenerator<Fr, ()>>,
+    // Option<TraceWitness<Fr>>,
```
leolara, line 37: @qwang98 remove commented out line

src/frontend/python/mod.rs
```diff
@@ -46,25 +58,93 @@ pub fn chiquito_ast_to_halo2(ast_json: &str) -> UUID {
...
+    assignment_generator: AssignmentGenerator<Fr, ()>,
+    rust_id: UUID,
+) {
+    CIRCUIT_MAP.with(|circuit_map| {
+        let mut circuit_map = circuit_map.borrow_mut();
+        let circuit_map_store = circuit_map.get_mut(&rust_id).unwrap();
+        circuit_map_store.2 = Some(assignment_generator);
+    });
+
+    println!("Added AssignmentGenerator to rust_id: {:?}", rust_id);
```
leolara, line 79: @qwang98 remove println

leolara (approved):
Just a few comments, but it is approved

examples/mimc7.py
qwang98: Fixed. I replaced the imports parameter for `Circuit` as `**kwargs`, so that the user can pass in arbitrarily named parameters for imports.

examples/mimc7.py
qwang98: Fixed.

src/frontend/dsl/mod.rs
qwang98: Fixed.

src/frontend/python/mod.rs
qwang98: Fixed.

src/frontend/python/mod.rs
qwang98: Fixed.

Issue #104, opened 2023-08-23
Add unit tests to FixedGenContext


leolara:
Here: https://github.com/privacy-scaling-explorations/chiquito/blob/main/src/wit_gen.rs

saeyoon17:
Hi, I would like to work on this issue. Can I claim it?

leolara:
@saeyoon17 assigned

qwang98:
Close as complete

PR #127, opened 2023-09-12
add unit test code for FixedGenContext - Issue #104

saeyoon17:
Hi, I wrote two unit tests regarding FixedGenContext as suggested in Issue #104.

- test_fixed_gen_context() checks the length of assignment is 1 given a single fixed column.
- test_fixed_gen_context_multiple() checks the length of assignment is 2 given two column assignments.

Please let me know if anything else is required. :)


leolara:
@saeyoon17 I think we should test that the values are correct, not just the length

saeyoon17:
@leolara I changed the test so that it now checks the value. Please tell me if there are any other modifications we should make. :)

leolara (approved):

leolara:
@saeyoon17 it is approved, does it show the button to merge to you?

saeyoon17:
@leolara No.. I think I only have the read access to this repository. This is how I see it:

image

qwang98:
@leolara  I think only maintainers are allowed to merge it. Any means we can change this in the repo settings?

PR #133, opened 2023-09-16
Automatic plaf instance - Issue #129

### LINKED ISSUE #129: Enable automatic generation of `instance` for Plaf backend
Halo2 mockprover::run requires an instance parameter, which Rust Chiquito automatically generates under the `instance` function in halo2.rs:
```
 pub fn instance(&self) -> Vec<Vec<F>> {
        if !self.compiled.circuit.exposed.is_empty() {
            if let Some(witness) = &self.witness {
                return vec![self.compiled.instance(witness)];
            }
        }
        Vec::new()
    }
```

which calls:
```
fn instance(&self, witness: &Assignments<F>) -> Vec<F> {
        let mut instance_values = Vec::new();
        for (column, rotation) in &self.circuit.exposed {
            let values = witness
                .get(column)
                .unwrap_or_else(|| panic!("exposed column not found: {}", column.annotation));

            if let Some(value) = values.get(*rotation as usize) {
                instance_values.push(*value);
            } else {
                panic!(
                    "assignment index out of bounds for column: {}",
                    column.annotation
                );
            }
        }
        instance_values
    }
```

An example of using the public `instance()` function in fibonacci.rs (or really just any example file that exposes public signal):
```
let prover = MockProver::<Fr>::run(7, &circuit, circuit.instance()).unwrap();
```

We want to enable the same for the Plaf backend. The final result should enable us to call `instance()` no plaf circuits as well:
```
let prover_plaf = MockProver::<Fr>::run(8, &plaf_circuit, plaf_circuit.instance()).unwrap();
```

Note that in Chiquito, the instance values are collected from the `Assignment` object indexed by the `instance` column.

More details to be filled for Plaf.

As a background, Plaf is a good visualization tool that checks whether polynomial constraints are identical among different Halo2 constructions (say between Chiquito and Halo2). It also has pretty print methods for witness and fixed assignments, that are good for debugging in general.

jae-cuz:
Resolves #129. Implemented automatic plaf instance generation.
`use chiquito::plonkish::backend::plaf::PlafInstance` would enable `plaf_circuit.instance()`.

Ready for review @leolara @qwang98

- Collect copy constraints btw witness_column & public_column, and index and offset of them
- `instance_with_offsets`: vector of pair `(witness value, public offset)`, where public offset indicates the order to be returned
- sort the vector `instance_with_offsets` with public offset
- `instance_values`: convert witness values in `instance_with_offsets` to field
- and return the instance vector

See along with example :
```
// when we use expose, plaf.copys is assigned like
[
    CopyC {
        columns: (
            Column {
                kind: Public,
                index: 0,
            },
            Column {
                kind: Witness,
                index: 1,
            },
        ),
        offsets: [
            (
                0,
                10,
            ),
        ],
    },
    CopyC {
        columns: (
            Column {
                kind: Public,
                index: 0,
            },
            Column {
                kind: Witness,
                index: 2,
            },
        ),
        offsets: [
            (
                1,
                10,
            ),
        ],
    },
]
// which means that values to be exposed are "witness[column 1][row 10]" and "witness[column 2][row 10]"
// in order 0, 1, respectively.
```


jae-cuz:
Tried generalizaiton with reference to https://github.com/Dhole/polyexen/blob/master/src/plaf/backends/halo2.rs#L346-L367

leolara:
So, actually this is better if we do in plaf codebase, because it is in their responsability boundary,. So I would send this as a PR there. But keep this open in case they don't accept it we can add it here.

leolara:
Please, paste here the link to the PR to plaf when you have sent it, so we can track if it gets merged.

jae-cuz:
Here is my PR to plaf : https://github.com/Dhole/polyexen/pull/3

I would make a new issue if it is merged :) Thanks

jae-cuz:
Move to #144.

Issue #102, opened 2023-08-23
Add unit tests to SimpleStepSelectorBuilder 


leolara:
Here: https://github.com/privacy-scaling-explorations/chiquito/blob/main/src/plonkish/compiler/step_selector.rs

mhh001:
Hi, Could I work on this issue?

leolara:
@mhh001 assigned

PR #134, opened 2023-09-18
Don't add column for SSB with only one step type

### LINKED ISSUE #95: SimpleStepSelectorBuilder case when there is only one step
The SimpleStepSelectorBuilder creates a column for each step type, that gets to be 1 in the row where a step instance of that type starts.

When there is only one step type we don't need any column, because all steps are of this step type.

Make SimpleStepSelectorBuilder to don't add any columns where there is a column, and create a StepSelector that works without any column.

000wan:
# Summary
It resolves #95.

It makes SimpleStepSelectorBuilder not add any step selector columns if there is only one step.
Further, I wrote a unit test for this (SSB with single step) and it might be helpful for #102.

# Test Results
Passes all tests with `cargo test`.
Runs Ok with `/examples/fibonacci.rs`, `/examples/mimc7.rs`, `/examples/poseidon.rs`
Witness display results of `cargo run --package chiquito --example fibonacci` changed:

### Before:
```
offset,srcm forward a,srcm forward b,srcm internal signal c,'step selector for fibo step'
0,1,1,2,1
1,1,2,3,1
2,2,3,5,1
3,3,5,8,1
4,5,8,13,1
5,8,13,21,1
6,13,21,34,1
7,21,34,55,1
8,34,55,89,1
9,55,89,144,1
10,89,144,233,1
```
### After:
```
offset,srcm forward a,srcm forward b,srcm internal signal c
0,1,1,2
1,1,2,3
2,2,3,5
3,3,5,8
4,5,8,13
5,8,13,21
6,13,21,34
7,21,34,55
8,34,55,89
9,55,89,144
10,89,144,233
```


000wan:
Looking forward to your reviews and comments!
Btw, I'm currently into this SSB code so it would be wonderful if there are other issues/features related to this that I can work on.

src/plonkish/compiler/step_selector.rs
```diff
@@ -74,6 +74,26 @@ impl StepSelectorBuilder for SimpleStepSelectorBuilder {
...
+                .selector_expr
+                .insert(step.uuid(), PolyExpr::Const(F::ONE));
+
+            selector
+                .selector_expr_not
+                .insert(step.uuid(), PolyExpr::Const(F::ZERO));
+
+            selector
+                .selector_assignment
+                .insert(step.uuid(), vec![(PolyExpr::Const(F::ONE), F::ONE)]);
```
leolara, line 91: @000wan this looks not the more readable way. I would make get_selector_assignment to return an Option, and then the code in assignment generator, could ignore this if None is returned.

000wan: I agree with that. Now I modified `selector_assignment` to have an Option value. Could you check this again? @leolara

src/plonkish/compiler/step_selector.rs
```diff
@@ -16,7 +16,7 @@ pub type SelectorAssignment<F> = (PolyExpr<F>, F);
 pub struct StepSelector<F> {
     pub selector_expr: HashMap<StepTypeUUID, PolyExpr<F>>,
     pub selector_expr_not: HashMap<StepTypeUUID, PolyExpr<F>>,
-    pub selector_assignment: HashMap<StepTypeUUID, Vec<SelectorAssignment<F>>>,
+    pub selector_assignment: HashMap<StepTypeUUID, Option<Vec<SelectorAssignment<F>>>>,
```
000wan, line 19: I made this Option in order to distinguish the cases where the uuid is invalid, or the uuid is valid but has no assignments.

leolara:
@000wan we are in a hacker house mentoring, that is why we have not answered in so long. This PR will get merged for sure. Let me get back to you

000wan:
I didn't know you were busy.. then I'll wait for it

leolara:
@000wan so the only problem now is that if the step has more than one row, then it will be active also for the rows that is not the first row of the step. The best way to solve this is to make `q_enable` to be only one at the beginning of each step. this change of course, only if the number of steps is one.

So currently q_enable is 1 for all rows. if we have only one step with row height 3, then q_enable should be 1,0,0,1,0,0,1,0,0,.. and so on

000wan:
Then doesn't it require a new column for the step?

PR #135, opened 2023-09-19
Make MaxWidthCellManager work with fixed signals - Issue #97

### LINKED ISSUE #97: Make MaxWidthCellManager work with fixed signals
Currently only SingleRowCellManager can handle fixed signals. Make it work on MaxWidthCellManager.

Because the fixed signals use different columns than the other signals, we need a new parameter called `max_width_fixed`.

atsupontio:
close #97 

@leolara Ready for review. I added `max_width_fixed` and renamed `max_width` to `max_width_advice`.

And, I fixed `examples/poseidon.rs` to use one `CompilerConfig`.

Please let me know if anything else is required!


src/plonkish/compiler/cell_manager.rs
```diff
@@ -231,24 +231,26 @@ impl CellManager for SingleRowCellManager {
...
+            max_width_fixed,
             same_height,
         }
     }
 }
 
 impl CellManager for MaxWidthCellManager {
     fn place<F>(&self, unit: &mut CompilationUnit<F>) {
-        if (!unit.shared_signals.is_empty() || !unit.fixed_signals.is_empty()) && !self.same_height
+        if (!unit.shared_signals.is_empty()) && !self.same_height
```
leolara, line 251: @atsupontio this must be kept. Fixed signals should not work when same_height is disabled

src/plonkish/compiler/cell_manager.rs
```diff
@@ -299,6 +301,48 @@ impl CellManager for MaxWidthCellManager {
             forward_signal_row
         } as u32;
 
+
+
+        let mut fixed_signal_column: usize = 0;
+        let mut fixed_signal_row: usize = 0;
+
+        for fixed_signal in unit.fixed_signals.iter() {
+            let column = if placement.columns.len() <= fixed_signal_column {
```
leolara, line 310: @atsupontio you should add some tests, this will not work. Because placement.columns contains both advice and fixed signals, if there are both, for example when `fixed_signal_column == 0` but there are already columns that are advice the fixed column will not be created and the wrong column will be picked up

src/plonkish/compiler/cell_manager.rs
```diff
@@ -299,6 +301,48 @@ impl CellManager for MaxWidthCellManager {
...
+            );
+
+            fixed_signal_column += 1;
+            if fixed_signal_column >= self.max_width_fixed {
+                fixed_signal_column = 0;
+                fixed_signal_row += 1;
+            }
+        }
+
+        placement.base_height = if fixed_signal_column != 0 {
```
leolara, line 339: @atsupontio what if there is more rows in the advice than in the fixed?

examples/poseidon.rs
```diff
@@ -661,25 +661,22 @@ fn poseidon_super_circuit<F: PrimeField + Eq + Hash>(
     lens: Lens,
 ) -> SuperCircuit<F, ValuesAndLens<F>> {
     super_circuit::<F, ValuesAndLens<F>, _>("poseidon", |ctx| {
-        let single_config = config(SingleRowCellManager {}, SimpleStepSelectorBuilder {});
+        let config = config(MaxWidthCellManager::new(2, 2, true), SimpleStepSelectorBuilder {});
```
leolara, line 664: @atsupontio I think it would be better if run the example with both Cell Managers

src/plonkish/compiler/cell_manager.rs
```diff
@@ -299,6 +301,59 @@ impl CellManager for MaxWidthCellManager {
             forward_signal_row
         } as u32;
 
+
+
+        let mut fixed_signal_column: usize = 0;
+        let mut fixed_signal_row: usize = 0;
+        let mut column_pos = self.max_width_advice + fixed_signal_column;
```
leolara, line 308: @atsupontio we might not have used all the max_width_advice columns, for example max_width_advice can be 3, but there be only 2 forwards.

leolara: perhaps column pos is equivalent to the (number of actual advice columns inserted (placement.columns.len() before adding fixed columns)) + fixed_signal_column; so this variable is not needed.

I think the cleaner solution is:

Be adding for advice and fixed, the columns to different arrays, and after all have been added then merge both arrays into placement.columns.

examples/poseidon.rs
atsupontio: @leolara Thank you!
I'll fix it!

src/plonkish/compiler/cell_manager.rs
atsupontio: @leolara Thanks for your great advice and ideas !

src/plonkish/compiler/cell_manager.rs
```diff
@@ -262,20 +264,21 @@ impl CellManager for MaxWidthCellManager {
 
         let mut forward_signal_column: usize = 0;
         let mut forward_signal_row: usize = 0;
+        let mut forward_columns: Vec<Column> = Vec::new();
 
         for forward_signal in unit.forward_signals.iter() {
-            let column = if placement.columns.len() <= forward_signal_column {
+            let column = if placement.forward.len() <= forward_signal_column {
```
leolara, line 270: @atsupontio you are comparing signals with columns here, the left side should be forward_columns.len()

src/plonkish/compiler/cell_manager.rs
```diff
@@ -299,6 +302,61 @@ impl CellManager for MaxWidthCellManager {
...
 
+
+
+        let mut fixed_signal_column: usize = 0;
+        let mut fixed_signal_row: usize = 0;
+        let mut fixed_columns: Vec<Column> = Vec::new();
+
+
+        for fixed_signal in unit.fixed_signals.iter() {
+            let column = if placement.fixed.len() <= fixed_signal_column {
```
leolara, line 313: @atsupontio the left side should be fixed_columns.len()

src/plonkish/compiler/cell_manager.rs
```diff
@@ -299,6 +302,61 @@ impl CellManager for MaxWidthCellManager {
...
+            );
+
+            fixed_signal_column += 1;
+            if fixed_signal_column >= self.max_width_fixed {
+                fixed_signal_column = 0;
+                fixed_signal_row += 1;
+            }
+        }
+
+        placement.base_height = if fixed_signal_row > forward_signal_row {
```
leolara, line 342: @atsupontio you can assign to a new variable max of fixed_signal_row, forward_signal_row and use the result in one expression

src/plonkish/compiler/cell_manager.rs
```diff
@@ -262,20 +264,21 @@ impl CellManager for MaxWidthCellManager {
 
         let mut forward_signal_column: usize = 0;
         let mut forward_signal_row: usize = 0;
+        let mut forward_columns: Vec<Column> = Vec::new();
```
leolara, line 267: @atsupontio actually these are advice_colums, not forward_columns

leolara: Advice columns, are shared by forward, shared and internal signals

src/plonkish/compiler/cell_manager.rs
leolara: advice_columns.len()

src/plonkish/compiler/cell_manager.rs
```diff
@@ -515,4 +576,102 @@ mod tests {
...
+
+        let step1 = Rc::new(step1);
+        let step2 = Rc::new(step2);
+
+        unit.step_types.insert(1500, Rc::clone(&step1));
+        unit.step_types.insert(1501, Rc::clone(&step2));
+
+        // forward signals: a, b; step1 internal: c1, d, e; step2 internal c2
+
+        let cm = MaxWidthCellManager {
```
leolara, line 610: @atsupontio you should this test with different combinations of   max_width_advice and  max_width_fixed.

Including corner cases, like 0 and 1.

leolara: If there are some repetation in the code use helper functions in the test module

src/plonkish/compiler/cell_manager.rs
atsupontio: @leolara Thank you for your comment!
Is it possible to represent it concisely with the following code?

```
let mut max_row = std::cmp::max(fixed_signal_row, forward_signal_row);
placement.base_height = if (fixed_signal_column != 0 && forward_signal_column != 0) {
    max_row as u32 + 1
} else {
    max_row as u32
};
placement.base_height = max_row;
```

leolara: In that code, you are setting max_row as mut, but you are not mutating it, also you are setting placement.base_height twice. Also, why are you checking both things for zero? If both are zero, max_row will be zero.

atsupontio: @leolara Since we don't know whether `fixed_signal_row` or `forward_signal_row` will be assigned to `max_row`, we need to check both `xx_signal_column`, don't we?
The `xx_signal_column` could be 0 even if the `xx_signal_row` is not 0.

PR #136, opened 2023-09-20
docs: signals should be marked as 'n.a.' in the padding columns

e3243eric:
### What was wrong?
The `Exposing Signals` table signal `b` is always 34 when padding.

### How can it be fixed?
I think the signal `b` should be `n.a.` when padding.


jae-cuz:
Hi @e3243eric,
have you read this part? I suppose b should be copied through the end.
https://github.com/privacy-scaling-explorations/chiquito/blob/fdc0d7fd4867c3117ad78347ea40ae66b5827d29/tutorials/tutorial_pt3_ch5.ipynb#L90

e3243eric:
Oh, I see. So this is designed to illustrate padding, not based on the Fibonacci sequence as I initially thought. Thanks @jae-cuz  for coming back so quickly.

PR #138, opened 2023-09-25
feat: add CompilerConfigBuilder

### LINKED ISSUE #109: CompilerConfigBuilder
Create a builder patter struct that allows to give some options and will create a compiler config for us. It uses a fluid interface, each method returns a `&mut` of `self`.

The interface is this:

`config_builder().max_width(10).build()` will return a config with MaxWidthCellManager{max_height, same_height: true}, SimpleStepSelectorBuilder

`config_builder().max_height(1).build()` will return a config with MaxWidthCellManager, SimpleStepSelectorBuilder

Things that have no Cell Manager should throw a panic.

Like:

`config_builder().max_width(10).max_height(2).build()`

cwkang1998:
Implementation for a compiler config builder, currently attempted to implement according to a typed builder approach.

Quite new at rust, so if I am doing something terribly wrong please do point them out and I'll work on it!
Adding test in the next commit, feel free to review before that so that I could make the required modification as well.

Closes #109

### DIFF
--- a/src/plonkish/compiler/mod.rs
+++ b/src/plonkish/compiler/mod.rs
@@ -11,9 +11,15 @@ use crate::{
 use std::{hash::Hash, rc::Rc};
 
 use cell_manager::{CellManager, SignalPlacement};
+use halo2_proofs::plonk::{Advice, Column as Halo2Column};
 use step_selector::StepSelectorBuilder;
 
 use unit::CompilationUnit;
+
+use self::{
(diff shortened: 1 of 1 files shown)


leolara:
This is very different from the requirements

cwkang1998:
@leolara Can you point out what I should change to adhere to the requirements? I can attempt another implementation of this if you don't mind. 

If you have an example for the implementation you want can you share it so that I may reference it as well? Thanks in advance!

Issue #96, opened 2023-08-23
Make MaxWidthCellManager work with shared signals


leolara:
Currently MaxWidthCellManager does not work with shared signals (SingleRowCellManager does).

Make them work.

What is a shared signal?

A shared signal is a signal that is shared by all the steps, and you can do rotation arbitrarily between all the steps, but they only work when the circuit has all the steps with the same height. Forward signals are also shared between steps, but you can only do a rotation to the next step, however they work on circuits with steps of different height.

e3243eric:
I would like to take this on. I've looked into SingleRowCellManager and MaxWidthCellManager. Is there a difference between them, where SingleRowCellManager only has one fixed height, while MaxWidthCellManager allows for multiple heights with a maximum size?

leolara:
Yes, that is the difference

leolara:
@e3243eric assigned

rutefig:
@e3243eric are you working on this? Otherwise can I take it?

PR #140, opened 2023-09-27
feat: make MaxWidthCellManager work with shared signals

e3243eric:
Close Issue #96.

I added the shared signals into `MaxWidthCellManager` by referring to `forward_signal`. I would like to know if this is the correct approach, and I am willing to write some tests. Is there anything I need to be aware of? Any suggestions for testing would be great.


e3243eric:
Hello @leolara, ready for review.

leolara:
This must include tests, the test should check a circuit with shared signals only, with forward signals only, and a circuit with both types of signals. All test should have a version with internal signals and another without.

PR #141, opened 2023-09-28
MiMC7 Jupyter Notebook

qwang98:
@leolara Ready for review.

All MiMC7 notebook files are `part4` of the tutorial, with 4 chapters, the first 3 of which is fully written out (just because they sort of came out naturally rather than to an outline first, so I thought this actually saved more time), and the last of which is an actual outline with `// TODOs`.

Let me know your thoughts!


examples/mimc7_pt2.py
qwang98, line 1: This file is just for easier edit to the tutorial, so that I can modify the code here before pasting to Jupyter. It will disappear eventually and won't be merged to main.

leolara (approved):

leolara:
I like it, let's merge this and get the TODOs merged by 3rd Oct.

PR #144, opened 2023-10-02
Use plaf automatic instance

### LINKED ISSUE #129: Enable automatic generation of `instance` for Plaf backend

jae-cuz:
Resolves #129. @leolara @qwang98 Ready for review.

Automatic instance is now enabled in plaf codebase,(https://github.com/Dhole/polyexen/pull/3) and this PR would apply that change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ halo2_proofs = { git = "https://github.com/privacy-scaling-explorations/halo2.gi
 halo2curves = { git = 'https://github.com/privacy-scaling-explorations/halo2curves', tag = "0.3.2", features = [
     "derive_serde",
 ] }
-polyexen = { git = "https://github.com/Dhole/polyexen.git", rev = "67148a8aabb000a61a169b2533b2526c5b76c42f" }
+polyexen = { git = "https://github.com/Dhole/polyexen.git", rev = "4d128ad2ebd0094160ea77e30fb9ce56abb854e0" }
 num-bigint = { version = "0.4", features = ["rand"] }
 uuid = { version = "1.4.0", features = ["v1", "rng"] }
 serde = { version = "1.0", features = ["derive"] }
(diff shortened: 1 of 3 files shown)


leolara:
I close this, as it is merged into plaf

jae-cuz:
Hi @leolara
this is separate PR from #133. This PR only applies change in plaf. Would you check it agian?

leolara (approved):

PR #146, opened 2023-10-11
Fixed LHS auto typing conversion for Expr arithmetic

qwang98:
Now Expr arithmetic doesn't require LHS to be Expr but can be ToExpr as well.

Thanks to Sandro for pointing out this error.


qwang98:
Simple changes ready for review.

leolara (approved):

PR #152, opened 2023-10-15
Automatic signals generation

leolara:
This is useful for:
 + Arbitrary boolean expressions
 + Multiplicative inverse elimination
 + Automatic witness generation in general
 + Conversion to lower degree polynomial identities


src/frontend/dsl/mod.rs
```diff
@@ -310,6 +310,18 @@ impl<'a, F> StepTypeSetupContext<'a, F> {
     }
 }
 
+impl<'a, F: Eq + PartialEq + Hash + Debug + Clone> StepTypeSetupContext<'a, F> {
+    pub fn auto(&mut self, signal: Queriable<F>, expr: ASTExpr<F>) {
+        self.step_type.auto_signals.insert(signal, expr);
+    }
+
```
qwang98, line 317: Will `auto` ever be used alone? If not, can make it `fn`.

src/plonkish/backend/halo2.rs
```diff
@@ -30,6 +30,10 @@ impl<T: Field + From<u64>> ChiquitoField for T {
     fn mi(&self) -> Self {
         self.invert().unwrap_or(Self::ZERO)
     }
+
+    fn pow<S: AsRef<[u64]>>(&self, exp: S) -> Self {
+        Field::pow(self, exp)
+    }
```
qwang98, line 36: Don't see these called any where.

src/wit_gen.rs
```diff
@@ -121,6 +122,84 @@ impl<F: Default, TraceArgs> TraceGenerator<F, TraceArgs> {
...
+                .map(|s| s.clone())
+                .collect::<Vec<Queriable<F>>>()
+                .clone();
+
+            // in each round at least one new signal should be assigned
+            if pending.len() == pending_amount {
+                panic!("cannot infer some auto signals")
+            }
+            pending_amount = pending.len()
+        }
```
qwang98, line 199: Is my understanding correct that each while loop attempts to assign as many pending signals as possible given the existing assignments? Then in the next loop it assigns additional signals that might depend on the assignment of the prior loop?

Is there a way to make the `panic!` message more specific that shows which signal(s) cannot be inferred?

src/wit_gen.rs
```diff
@@ -228,4 +307,121 @@ mod tests {
...
+            Some(&Fr::from(3))
+        );
+        assert_eq!(
+            witness.step_instances[1].assignments.get(&b),
+            Some(&Fr::from(2))
+        );
+        assert_eq!(
+            witness.step_instances[1].assignments.get(&c),
+            Some(&Fr::ONE)
+        );
```
qwang98, line 425: Looks like this example is trying to test that if there's an existing assignment for `c = 1`, then the auto-generator won't mask it with inferred value `c = 3`.

I think you called `assert_eq!` twice on the same assignment `witness.step_instances[1].assignments.get(&b)`, so there's probably something else you are trying to assert.

src/frontend/dsl/mod.rs
```diff
@@ -310,6 +310,18 @@ impl<'a, F> StepTypeSetupContext<'a, F> {
...
+    pub fn auto_eq(&mut self, signal: Queriable<F>, expr: ASTExpr<F>) {
+        self.auto(signal.clone(), expr.clone());
+
+        self.constr(eq(signal, expr));
+    }
```
qwang98, line 322: Could we also add `auto_transition`? Not sure how much harder it would be with rotations.

qwang98 (changes requested):
Overall, this PR allows the `auto_eq` feature which automatically assigns the left hand side queriable of a regular constraint. I'm good with the implementation of this feature and left a couple comments on some potential improvements. Let me know if you agree!

Looking forward, there are more things we can do:
1. Allow auto witness generation for constraints other than `eq`.
2. Allow auto witness generation for transition constraints and signals with rotation as well.
3. Similar to PIL, we could eventually enable an auto witness generation "mode" where instead of using the `auto_eq` API, the user only needs to provide the initial trace input and all signals in all step instances will be inferred, or return an error saying that not enough trace input was provided.

src/frontend/dsl/mod.rs
leolara: Yes, auto is like <-- in circom.

src/plonkish/backend/halo2.rs
leolara: This is to make the traits compatible

src/wit_gen.rs
leolara: Yes, but I was leaving that for a small task for new people

src/frontend/dsl/mod.rs
leolara: There cannot be auto_transition, because the infer signal cannot depend on signals of next steps. Or perhaps it is possible but it would be more complicated to evaluate, so we should leave it as a posible future feature. It is not trivial, because there can be a chain of `nexts` in the auto signals

src/wit_gen.rs
leolara: This was an error in the tests, the duplicated assert_eq!, it is fixed now

qwang98: LGTM.

src/frontend/dsl/mod.rs
qwang98: Understood!

qwang98 (approved):
LGTM with additional features to be implemented in future PRs.

qwang98:
@leolara Good to go after fixing the conflicts.

PR #154, opened 2023-10-16
Feature/optimise step selector builder

rutefig:
### Description
- **Goal:** is to optimise the Step Selector Builder to have Log2(n) Columns per n Step Types instead of one Column per each Step Type.
- **Current Implementation:** implemented a new type of Selector and implemented the build function for it

- **Explanation:** If we use binary representation for the step types instead of using the actual columns we will get an optimised version for the Selector Builder since we can get the total of columns from the following expression
```math
n\_cols = \lceil \log_2(n\_step\_types + 1) \rceil
```
We need take into consideration that we also need to save one binary value for the case when there is no step type, because of this we do the math with \(n\_step\_types + 1\).

 ### What is missing
- [x] Ensure compatibility with the backend
- [x] Unit tests


leolara:
4 steps require 3 columns, because 00 is not used. Please add a test with 4 steps that checks that 3 columns are used.

src/plonkish/compiler/step_selector.rs
```diff
@@ -197,3 +259,137 @@ fn other_step_type<F>(unit: &CompilationUnit<F>, uuid: UUID) -> Option<Rc<StepTy
...
+                    Uuid::now_v1(&[1, 2, 3, 4, 5, 6]).as_u128(),
+                    format!("StepType{}", i),
+                )),
+            );
+        }
+
+        builder.build(&mut unit);
+
+        // Assert number of columns
+        let n_cols = (n_step_types as f64).log2().ceil() as usize;
```
leolara, line 386: +1 missing

src/plonkish/compiler/step_selector.rs
```diff
@@ -197,3 +259,137 @@ fn other_step_type<F>(unit: &CompilationUnit<F>, uuid: UUID) -> Option<Rc<StepTy
...
+            )),
+        );
+
+        builder.build(&mut unit);
+
+        // Assert number of columns
+        assert_eq!(unit.columns.len(), 3);
+        assert_eq!(unit.selector.columns.len(), 3);
+
+        // TODO: Assert expressions
```
leolara, line 358: @rutefig do TODOs

leolara:
Rusfmt problems

leolara:
I think you test the expresions and the assignments, the expressions can be converted to a string with `print!(":#?", expr) ` and then compared to a string @rutefig

src/plonkish/compiler/step_selector.rs
```diff
@@ -188,6 +188,69 @@ impl StepSelectorBuilder for TwoStepsSelectorBuilder {
...
+                    combined_expr_not = combined_expr_not
+                        * (PolyExpr::Const(F::ONE) - column.query(0, format!("Column {}", i)));
+                    assignments.push((column.query(0, format!("Column {}", i)), F::ONE));
+                } else {
+                    combined_expr = combined_expr
+                        * (PolyExpr::Const(F::ONE) - column.query(0, format!("Column {}", i)));
+                    combined_expr_not =
+                        combined_expr_not * column.query(0, format!("Column {}", i));
+                }
+            }
```
leolara, line 235: @rutefig I think the right combined_expr_not is always (1-combined_expr)

leolara (approved):

rutefig:
@leolara need approval here, the previous run failed on the Clippy so I committed a fix for that

leolara (approved):

PR #155, opened 2023-10-17
User defined parameter k for halo2 mock prover

qwang98:
Ready for review. @leolara

### DIFF
--- a/src/frontend/python/mod.rs
+++ b/src/frontend/python/mod.rs
@@ -83,6 +83,7 @@ fn add_assignment_generator_to_rust_id(
 pub fn chiquito_super_circuit_halo2_mock_prover(
     rust_ids: Vec<UUID>,
     super_witness: HashMap<UUID, &str>,
+    k: usize,
 ) {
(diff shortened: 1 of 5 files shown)


src/frontend/python/mod.rs
```diff
@@ -1864,7 +1866,11 @@ fn super_circuit_halo2_mock_prover(rust_ids: &PyList, super_witness: &PyDict) {
         })
         .collect::<HashMap<u128, &str>>();
 
-    chiquito_super_circuit_halo2_mock_prover(uuids, super_witness)
+    chiquito_super_circuit_halo2_mock_prover(
+        uuids,
+        super_witness,
+        k.extract().expect("PyLong convertion failed."),
```
leolara, line 1872: conversion?

leolara (approved):

src/frontend/python/mod.rs
qwang98: Yes. It's type conversion between Python and Rust. PyLong is basically int.

PR #165, opened 2023-10-28
PIL Backend with PIL IR

qwang98:
Ready for review. @leolara

# High-Level Approach
Convert Chiquito AST to PIL IR and finally to PIL code.

# Future TODOs
- Add support for trace inputs to PIL. Currently, PIL supports user input via the command line or Rust APIs for the command line. However, it only supports input for one variable (column). Because Chiquito circuit might provide inputs for multiple variables via trace (e.g. the MiMC7 circuit), we can't use use PIL's Rust API as a generic solution for providing inputs. The only solution is to input statements in the format of `ISFIRST * ([[input_variable]] - [[input_value]]) = 0`. An issue with this approach, however, is that our `trace` is too general and doesn't associate input values to signals, so we might need to create an alternative `trace` function or mode that does this. All of the above is due to the fact that PIL only supports automatic witness inference and cannot feed in external witness.
- Add Python PIL API support for super circuits. Currently we support single circuit in Python. Should be very do-able, just didn't get the time to code it up.


leolara:
We can solve the trace issue in the future.

leolara:
I think the structure should be:
plonkish/ir/pil for the IR data structure
plonkish/backend/pwdr_pil for the code that transform that IR into PWDR PIL

qwang98:
> I think the structure should be: plonkish/ir/pil for the IR data structure plonkish/backend/pwdr_pil for the code that transform that IR into PWDR PIL

Fixed! Ready for review again. @leolara

examples/fibonacci.rs
```diff
@@ -163,3 +173,17 @@ fn main() {
         }
     }
 }
+
+#[cfg(test)]
+#[test]
+fn test_pil_single_circuit_fibo() {
```
leolara, line 179: Does it make sense to add a test to an example? shouldn't it be in the main?

src/plonkish/backend/powdr_pil.rs
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    pil
+}
+
+// Generate PIL code given the IR of a single PIL circuit. Need to supply `annotations_map`,
+// which only need to contain annotations for the circuit's own components if we are converting
+// a standalone circuit. If we are converting a super circuit, `annotations_map` also needs to
+// contain annotations for components outside the circuit.
+fn generate_pil_given_annotations<F: Clone + Debug>(
```
leolara, line 119: @qwang98 this function is too long, please split it

src/plonkish/ir/pil.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub step_types: Vec<UUID>,
+    pub constraints: HashMap<UUID, Vec<Expr<F, Queriable<F>>>>, // step type UUID -> constraints
+    pub transitions: HashMap<UUID, Vec<Expr<F, Queriable<F>>>>,
+    pub lookups: HashMap<UUID, Vec<PilLookup<F>>>, // step type UUID -> lookups vector
+}
+
+type PilLookup<F> = Vec<(Expr<F, Queriable<F>>, Expr<F, Queriable<F>>)>;
+
+impl<F: Clone + Debug> PilCircuit<F> {
+    pub fn from_ast_and_witness<TraceArgs>(
```
leolara, line 37: @qwang98 this function is too long, please slit it.

leolara (changes requested):
Although currently each AST queriable correspond to a PIL column, that does not have to be true in the future. In the IR we should have our own Column type with information corresponding with the information of the IR columns

src/plonkish/ir/pil.rs
leolara: Converting AST to IR is the compiler role, this function is actually a compiler and should have its own module and be called PIL compiler.

leolara:
Given that PIL needs it own compiler IR and two backends (polygon and powdr), perhaps we should create a new directory structure for it:

`pil/{IR,compiler,backend}`

qwang98:
Sorry was a bit busy with school last week. Will take a look tomorrow!

examples/fibonacci.rs
qwang98: Agree and fixed!

src/plonkish/backend/powdr_pil.rs
qwang98: Fixed.

qwang98:
> Given that PIL needs it own compiler IR and two backends (polygon and powdr), perhaps we should create a new directory structure for it:
> 
> `pil/{IR,compiler,backend}`

Note that we only implemented Powdr PIL not Polygon PIL. Reasons are two fold:
1. We originally found that Powdr PIL supports arbitrary degree constraints, though we later found that arbitrary degree is only for Halo2, while eSTARK requires degree two constraints. These said, Polygon PIL only supports degree 2 (and likely R1CS only) constraints regardless.
2. Powdr PIL's column definitions in their syntax work better with Chiquito than Polygon PIL.

However, in case we also have the chance to implement a Polygon PIL backend in the future, I think this folder structure can be adopted. It'll also be good to explicitly show that we are doing Powdr PIL rather than Polygon PIL through the file names.

qwang98:
> Although currently each AST queriable correspond to a PIL column, that does not have to be true in the future. In the IR we should have our own Column type with information corresponding with the information of the IR columns

Yes, currently each queriable is its own column, similar to our simple cell manager model where each step instance is one single row (or can our simple cell manager also place multiple signals into the same column across different rows?)

The current implementation is not the most efficient in terms of saving the number of columns, and I agree that in the future, we could have our own PIL IR Column type.

However, I'd love to point out that PIL only supports rotation of one (similar to our `next` for forward signals), and I'm not sure if this will work with step instances with multiple rows though.

So we have two issues here:
1. Support simple cell manager and have # of columns == maximum # of signals for each step type. This should work.
2. Support multi row cell manager, which would be harder and require some research.

qwang98:
> Although currently each AST queriable correspond to a PIL column, that does not have to be true in the future. In the IR we should have our own Column type with information corresponding with the information of the IR columns

I opened issue #183 for this.

src/plonkish/ir/pil.rs
qwang98: Agreed and fixed.

qwang98:
Ready for review again :)

qwang98:
> Although currently each AST queriable correspond to a PIL column, that does not have to be true in the future. In the IR we should have our own Column type with information corresponding with the information of the IR columns

Opened issue #183 for this.

src/pil/compiler/powdr_pil.rs
```diff
@@ -0,0 +1,184 @@
...
+            transitions.insert(
+                step_type.uuid(),
+                step_type
+                    .transition_constraints
+                    .iter()
+                    .map(|transition| transition.expr.clone())
+                    .collect::<Vec<Expr<F, Queriable<F>>>>(),
+            );
+
+            // Note that there's no lookup table in PIL, so we only need to convert lookups.
```
leolara, line 165: @qwang98 I don't understand this comment, if there is no lookup table in PIL, we should not add them, give a panic if someone tries to compile a circuit with lookup to PIL

src/pil/ir/powdr_pil.rs
```diff
@@ -0,0 +1,29 @@
...
+    // circuit level
+    pub circuit_name: String,
+    pub num_steps: usize,
+
+    // circuit level - col witness
+    pub col_witness: Vec<UUID>, // internal signal NOT dedupped
+
+    // circuit level - col fixed
+    pub fixed_signals: HashMap<UUID, Vec<F>>, // fixed signal UUID -> fixed assignments vector
+    pub step_types_instantiations: HashMap<UUID, Vec<usize>>, /* step type UUID -> step
```
leolara, line 16: @qwang98 there is no steps in PIL, the PIL shouldn't have steps. The fact that this is here might mean that there is a translation that you are doing in the backend that should be done in the compiler

src/pil/ir/powdr_pil.rs
```diff
@@ -0,0 +1,29 @@
...
+                                               * instances
+                                               * vector {0, 1}^num_steps */
+    pub first_step: Option<UUID>,
```
leolara, line 19: @qwang98 same, there is no step in PIL

src/plonkish/ir/sc.rs
```diff
@@ -29,6 +31,30 @@ impl<F, MappingArgs> SuperCircuit<F, MappingArgs> {
     pub fn get_mapping(&self) -> MappingGenerator<F, MappingArgs> {
         self.mapping.clone()
     }
+
```
leolara, line 34: @qwang98   The IR from plonklish should not be used in PIL

src/pil/backend/powdr_pil.rs
```diff
@@ -0,0 +1,488 @@
...
+        }
+    }
+}
+
+fn generate_pil_step_types<F: Debug + Clone>(
+    pil: &mut String,
+    pil_ir: &PilCircuit<F>,
+    annotations_map: &HashMap<UUID, String>,
+) {
+    if !pil_ir.step_types.is_empty() {
```
leolara, line 283: If the step selector columns for PIL are normal advice or fixed columns, then they should be in the IR as normal PIL columns

src/pil/backend/powdr_pil.rs
```diff
@@ -0,0 +1,488 @@
...
+        for lookup in lookups {
+            let mut lookup_source: Vec<String> = Vec::new();
+            let mut lookup_destination: Vec<String> = Vec::new();
+            for (src, dest) in lookup {
+                lookup_source.push(convert_to_pil_expr_string(src.clone(), annotations_map));
+                lookup_destination.push(convert_to_pil_expr_string(dest.clone(), annotations_map));
+            }
+            // PIL lookups have the format of `step_type { lookup_sources } in {
+            // lookup_destinations }`.
+            writeln!(
```
leolara, line 385: @qwang98 Do we have or we don't have lookups in PIL?

src/pil/backend/powdr_pil.rs
```diff
@@ -0,0 +1,488 @@
...
+}
+
+// Convert queriable to PIL string recursively. Major differences are: 1. PIL rotations are in the
+// format of an apostrophe after the signal; 2. PIL only supports the next rotation, so there's no
+// previous or arbitrary rotation.
+fn convert_to_pil_queriable_string<F>(
+    query: Queriable<F>,
+    annotations_map: &HashMap<UUID, String>,
+) -> String {
+    match query {
```
leolara, line 447: We should have a column type already in PIL IR, this translation should be done in the compiler

src/pil/compiler/powdr_pil.rs
```diff
@@ -0,0 +1,184 @@
+use crate::{
```
leolara, line 1: @qwang98 the compiler is not specific for the POWDR PIL, it will be the same for the polygon PIL, so all this code should be moved to mod.rs

qwang98:
Create PIL Column and Queriable types, where the queriable rotation is bool. Software architecture fix.

src/pil/compiler/powdr_pil.rs
qwang98: PIL doesn't have lookup table as to store a set of destination columns. They just store the fixed columns separately. PIL still has lookup (which will be the "application" of the lookup table in the case of Chiquito). I deleted these comments to avoid confusion.

src/pil/ir/powdr_pil.rs
qwang98: Agreed that PIL shouldn't have steps, but we do need to express steps and step instantiations in PIL through fixed signals. For example: https://github.com/privacy-scaling-explorations/chiquito/blob/8803c3a47793a37ea7d039c63e68434d87d7d472/examples/mimc7.pil#L7-L10

I removed `step_types_instantiations` from PIL IR and converted all of them to `fixed_signals` in compiler instead.

Note that we still need `first_step` and `last_step` fields in PIL IR, although they are not PIL concepts. We need them to enforce things like `pragma_first_step` and `pragma_num_step` in PIL, which are enforced through constraints like the following example: https://github.com/privacy-scaling-explorations/chiquito/blob/8803c3a47793a37ea7d039c63e68434d87d7d472/examples/mimc7.pil#L43-L46

To your point, we could technically convert `first_step` and `last_step` to `constraints` in PIL IR, but that'll require modifying the annotation HashMap and the layers that use this data structure and can be quite error prone, so I elected to keep `first_step` and `last_step` in PIL IR.

src/pil/backend/powdr_pil.rs
qwang98: Agreed and fixed.

src/plonkish/ir/sc.rs
qwang98: I agree. I appended the sub circuit asts (now SBPIR) to the super circuit IR and that's why I took our IR as an input (so that I can access the sub circui asts appended). I think this is a quite clumsy work around and I'll instead take the sub circuit asts directly.

src/pil/backend/powdr_pil.rs
qwang98: We do have lookups (source to destination columns) in PIL but not lookup tables (destination columns only).

src/pil/backend/powdr_pil.rs
qwang98: I see your point. I can create a column and query system in PIL IR, though it'll be almost identical to our Queriable, maybe except for the rotation part.

src/pil/compiler/powdr_pil.rs
qwang98: I agree. The real difference comes from how we convert IR to backend PIL string. I think there could be a chance someone writes a Polygon PIL converter from our PIL IR to Polygon PIL string.

src/pil/ir/powdr_pil.rs
qwang98: I see your point. I'll try to convert them to constraints in PIL IR.

qwang98:
All comments addressed. Ready for review again. @leolara

leolara:
@qwang98 

What is the situation currently with the witness in PIL, can we pass some input to the proof generation, even most of the witness is created by PIL?

qwang98:
> @qwang98
> 
> What is the situation currently with the witness in PIL, can we pass some input to the proof generation, even most of the witness is created by PIL?

Yes. PIL has auto witness generation and actually only auto witness generation. Witnesses are typically created by passing in inputs for the first row in the form of a constraint, e.g. ISFIRST * (2 - input) = 0. However, as I explored before, there's no way of converting Chiquito TraceArgs automatically to such constraint, because Chiquito TraceArgs have no string names for them, which is required for creating such constraints. As we discussed last time (again a while back), the plan was to have the user create these constraints themselves.

leolara (approved):
I think we should consider this as experimental as the proof only works for one witness.


We will have to find a way in the future to pass the first row to the circuit to generate different witnesses.

leonardoalt:
btw powdr now accepts partial external witnesses as well. The partial comes from "a subset of the witness columns", but each column must be full, not sure whether that helps.

Otherwise, you can also use public/private inputs to set specific witnesses (also not sure if that solves your problem).

PR #167, opened 2023-10-31
Added pil examples

qwang98:
@leolara Ready for review.

These both ran successfully in Powdr.


leolara:
Why should we have examples in PIL?

leolara:
These are PIL code, that for us is like an IR, in the end we want their backend result, so not relevant as an example. As our users don't have to write PIL

PR #173, opened 2023-11-02
Python frontend lookup fix

qwang98:
Two issues fixed:
1. Serde deserializer failed to store the parsed Lookup in Rust AST.
2. Python frontend uses the single circuit compile function rather than the by phase compile functions for super circuits.


qwang98:
@leolara Ready for review.

leolara (approved):

PR #175, opened 2023-11-04
Keep `witness` argument immutable on `halo2_mock_prover`

### LINKED ISSUE #170: Mock prover modifies the given witness
When using the `halo2_mock_prover`, the given `witness` is modified internally, and the data structure of it changes.

It doesn't represent a big problem, but I think it's obscure. It would feel more natural if, any extra data the prover wants to expose, it is returned by the `halo2_mock_prover`, instead of altering the arguments.

So instead of 
```
super_circuit.halo2_mock_prover(witness)
```

we could have 
```
result = super_circuit.halo2_mock_prover(witness)
```

### Before prover
https://github.com/sraver/chiquito/blob/bd4402000ace7813d09326c938819d5d6a886943/examples/lookup_table.py#L141
```
TraceWitness(
        step_instances={
                StepInstance(
                        step_type_uuid=230416691034002696322218178563671722506,
                        assignments={
                                a = 1,
                                b = 2
                        },
                ),
                StepInstance(
                        step_type_uuid=230416691034002696322218178563671722506,
                        assignments={
                                a = 2,
                                b = 3
                        },
                ),
                ...
        },
)
```

### After prover
https://github.com/sraver/chiquito/blob/bd4402000ace7813d09326c938819d5d6a886943/examples/lookup_table.py#L146
```
{
    "step_instances": [
        {
            "step_type_uuid": 282599730103441542450410585628967832074,
            "assignments": {
                "282599718694586140393929509483578001930": [
                    {
                        "Forward": [
                            {
                                "id": 282599718694586140393929509483578001930,
                                "phase": 0,
                                "annotation": "a"
                            },
                            false
                        ]
                    },
                    [
                        12436184717236109307,
                        3962172157175319849,
                        7381016538464732718,
                        1011752739694698287
                    ]
                ],
                "282599726062805254222938657032457816586": [
                    {
                        "Forward": [
                            {
                                "id": 282599726062805254222938657032457816586,
                                "phase": 0,
                                "annotation": "b"
                            },
                            false
                        ]
                    },
                    [
                        6425625360762666998,
                        7924344314350639699,
                        14762033076929465436,
                        2023505479389396574
                    ]
                ]
            }
        },
        {
            "step_type_uuid": 282599730103441542450410585628967832074,
            "assignments": {
                "282599718694586140393929509483578001930": [
                    {
                        "Forward": [
                            {
                                "id": 282599718694586140393929509483578001930,
                                "phase": 0,
                                "annotation": "a"
                            },
                            false
                        ]
                    },
                    [
                        6425625360762666998,
                        7924344314350639699,
                        14762033076929465436,
                        2023505479389396574
                    ]
                ],
                "282599726062805254222938657032457816586": [
                    {
                        "Forward": [
                            {
                                "id": 282599726062805254222938657032457816586,
                                "phase": 0,
                                "annotation": "b"
                            },
                            false
                        ]
                    },
                    [
                        415066004289224689,
                        11886516471525959549,
                        3696305541684646538,
                        3035258219084094862
                    ]
                ]
            }
        },
        ...
    ]
}

```

Not sure if this is really an issue, I'm opening the ticket more to create conversation about it. I believe keeping the `witness` immutable and returning whatever data is required would be easier to understand.

sraver:
Closes #170 

---

On the SuperCircuit's `halo2_mock_prover` method, we are currently overriding the `super_witness` dict object.
Initially it holds the `TraceWitness` objects of each internal circuit, but after the loop it holds the JSON representation of the witness of the circuit.

Since the elements in the dict are indexed by their `rust_id`, after the loop all the elements have been replaced, leaving the caller of `halo2_mock_prover` with a "modified" `witness`.

It's not a big problem, but it's confusing sometimes when writing tests, if you want to verify some of the assignments, because the order of lines becomes relevant.

I didn't add the return of the `witness_json` because so far I've seen no use for it. Maybe it could be interesting.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "chiquito"
-version = "0.1.2023110200"
+version = "0.1.2023110700"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 authors = ["Leo Lara <leo@leolara.me>"]
--- a/src/frontend/python/chiquito/dsl.py
+++ b/src/frontend/python/chiquito/dsl.py
@@ -73,15 +73,15 @@ class SuperCircuit:
     def halo2_mock_prover(
         self: SuperCircuit, super_witness: Dict[int, TraceWitness], k: int = 16
     ):
+        witness_json = {}
         for rust_id, witness in super_witness.items():
             if rust_id not in self.ast.sub_circuits:
                 raise ValueError(
                     f"SuperCircuit.halo2_mock_prover(): TraceWitness with rust_id {rust_id} not found in sub_circuits."
                 )
-            witness_json: str = witness.get_witness_json()
-            super_witness[rust_id] = witness_json
+            witness_json[rust_id] = witness.get_witness_json()
         rust_chiquito.super_circuit_halo2_mock_prover(
-            list(self.ast.sub_circuits.keys()), super_witness, k
+            list(self.ast.sub_circuits.keys()), witness_json, k
         )
 
 


qwang98 (approved):
LGTM and thanks for suggesting this change! We don't need to return witness JSON as it's only used to serialize a Python object to be parsed by Rust.

qwang98:
@sraver I'd just request that you change the version number below to the format [[0.1.YYYYMMDDVV]] where [[VV]] can be 00 if it's the first pull request we merge to the main on a day. Only this will trigger the update for the published Python package.

https://github.com/privacy-scaling-explorations/chiquito/blob/main/Cargo.toml#L3

After that, feel free to merge. :)

sraver:
@qwang98, just updated the version.

I can't merge tho, please proceed for me ^_^

PR #176, opened 2023-11-08
Add optional keyword arguments to `SuperCircuit`

sraver:
In some cases, we need to pass information into the Circuit/SuperCircuit, and for that, we can use optional keyword arguments sent through the constructor.

This PR adds this feature, already present in `Circuit`, into `SuperCircuit`.


qwang98 (approved):
LGTM as discussed :)

PR #178, opened 2023-11-18
Port Poseidon example to Python

### LINKED ISSUE #101: Example: Port poseidon to python front-end
We have it implemented in rust front-end. Do it in python front-end

sraver:
Closes #101 

---

Based on the Rust implementation, the same circuit is added using the Python front-end.

Constants are moved to a different file for easier code browsing.

An assertion is added when running the example, to make sure that the output is the same as the one when using the Rust front-end.


leolara:
Please review @10to4

10to4 (approved):

PR #184, opened 2023-11-26
HyperPlonk backend

qwang98:
Ready for review @leolara.

I asked Han about the best approach to testing, for which he suggested exposing functions from his repo, for which I did over my fork of his repo in its main branch: https://github.com/qwang98/plonkish.


src/plonkish/backend/hyperplonk.rs
```diff
@@ -0,0 +1,441 @@
...
+pub struct ChiquitoCircuit<F, TraceArgs> {
+    // TraceArgs is for trace_gen, which is not used here; we only care about placement
+    k: usize,
+    instances: Vec<Vec<F>>, /* outter vec has length 1, inner vec has length equal to number of
+                             * exposed signals */
+    circuit: Circuit<F>,
+    num_witness_polys: Vec<usize>,
+    advice_idx_in_phase: Vec<usize>,
+    assignment_generator: AssignmentGenerator<F, TraceArgs>,
+    trace_witness: TraceWitness<F>,
```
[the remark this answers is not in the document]
qwang98: Take out TraceWitness

src/plonkish/backend/hyperplonk.rs
```diff
@@ -0,0 +1,441 @@
...
+        let selector_assignments = self
+            .selector_uuids
+            .iter()
+            .map(|uuid| {
+                selector_assignments_unordered
+                    .get(&self.circuit.columns[column_idx(*uuid, &self.all_uuids)])
+                    .unwrap()
+                    .clone()
+            })
+            .collect::<Vec<Vec<F>>>();
```
qwang98, line 311: Remove these from the preprocessing phase as they can change between different witnesses.

src/plonkish/backend/hyperplonk.rs
```diff
@@ -0,0 +1,441 @@
...
+    circuit: Circuit<F>,
+    num_witness_polys: Vec<usize>,
+    advice_idx_in_phase: Vec<usize>,
+    assignment_generator: AssignmentGenerator<F, TraceArgs>,
+    trace_witness: TraceWitness<F>,
+    all_uuids: Vec<UUID>,      // the same order as self.circuit.columns
+    fixed_uuids: Vec<UUID>,    // the same order as self.circuit.columns
+    selector_uuids: Vec<UUID>, // the same order as self.circuit.columns
+    non_selector_advice_uuids: Vec<UUID>, // the same order as self.circuit.columns
+    non_selector_advice_uuids_by_phase: HashMap<usize, Vec<UUID>>,
```
qwang98, line 106: Remove the selector. Chiquito selector and non selector will both be advice. They are not fixed.

Replace with a HashMap as it's O(logN) while vector fine is O(N)

qwang98: Note that we need to find the index of the UUID within the collection and keep the order, but HashMap doesn't keep any order, so we still have to use a Vec

src/plonkish/backend/hyperplonk.rs
```diff
@@ -0,0 +1,441 @@
...
+    num_by_phase(phases)
+        .into_iter()
+        .scan(0, |state, num| {
+            let offset = *state;
+            *state += num;
+            Some(offset)
+        })
+        .collect()
+}
+
```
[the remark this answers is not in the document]
qwang98: Done

src/plonkish/backend/hyperplonk.rs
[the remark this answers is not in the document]
qwang98: Done

src/plonkish/backend/hyperplonk.rs
```diff
@@ -0,0 +1,441 @@
...
+pub struct ChiquitoCircuit<F, TraceArgs> {
```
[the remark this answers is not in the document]
qwang98: Done

src/plonkish/backend/hyperplonk.rs
qwang98: Done.

examples/fibonacci.rs
```diff
@@ -95,7 +100,7 @@ fn fibo_circuit<F: Field + From<u64> + Hash>() -> (Circuit<F>, Option<Assignment
             let mut a = 1;
             let mut b = 2;
 
-            for _i in 1..11 {
+            for _i in 1..16 {
```
qwang98, line 103: Hyperplonk requires 2^integer number of rows, but in practice this can be replaced by padding, so this shouldn't affect our example.

qwang98:
Note that build and clippy error on a dependency are caused by not using the latest nightly build. Please override or default to the lastest nightly when reviewing this PR.

examples/fibonacci.rs
```diff
@@ -162,4 +167,22 @@ fn main() {
...
+        backend,
+        halo2_curves::bn256::{Bn256, Fr},
+        pcs::{multilinear, univariate},
+    };
+    // get Chiquito ir
+    let (circuit, assignment_generator) = fibo_circuit::<Fr>();
+    // get assignments
+    let assignments = assignment_generator.unwrap().generate(());
+    // get hyperplonk circuit
+    let hyperplonk_circuit = ChiquitoHyperPlonkCircuit::new(4, circuit, assignments);
```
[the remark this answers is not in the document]
qwang98: I think that's actually how it was implemented in the prior iteration. I unresolved the relevant comments above, which are notes taken during a meeting we had. So back then it was suggested that I take an assignment directly instead.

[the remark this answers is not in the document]
qwang98: Yes. That's exactly what we have. See the interfaces here, which splits it into circuit and assignments:

https://github.com/privacy-scaling-explorations/chiquito/blob/e0e2c9c05d3ee7c0d49caf625a394b23189aff4a/src/plonkish/backend/hyperplonk.rs#L71-L86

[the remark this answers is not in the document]
qwang98: Do you mean that I should create a `new` function that doesn't take an assignment but another function that allows the user to add the assignment later?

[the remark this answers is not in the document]
qwang98: I separated this into two functions, one creating the circuit and the other adding assignments.

examples/fibonacci.rs
```diff
@@ -162,4 +167,23 @@ fn main() {
...
+        halo2_curves::bn256::{Bn256, Fr},
+        pcs::{multilinear, univariate},
+    };
+    // get Chiquito ir
+    let (circuit, assignment_generator) = fibo_circuit::<Fr>();
+    // get assignments
+    let assignments = assignment_generator.unwrap().generate(());
+    // get hyperplonk circuit
+    let mut hyperplonk_circuit = ChiquitoHyperPlonkCircuit::new(4, circuit);
+    hyperplonk_circuit.set_assignment(assignments);
```
leolara, line 184: @qwang98 so what happens if you don't set the assigments, the circuit gets compiled?

leolara: You need to be able to compile the circuit to hyperplonk, and then on the compiled circuit to be able to generate proof with the assignments. The assignments are needed in the proof generation phase, not for the compilation.

As you can see in the halo2 example.

qwang98: > @qwang98 so what happens if you don't set the assigments, the circuit gets compiled?

Yes. The circuit gets compiled even without the assignments. For example, constraints are converted to HyperPlonk format without needing assignments.

qwang98: > You need to be able to compile the circuit to hyperplonk, and then on the compiled circuit to be able to generate proof with the assignments. The assignments are needed in the proof generation phase, not for the compilation.
> 
> As you can see in the halo2 example.

This is what we have, but I believe your point is that only the proof generation function should take the assignment, no the compilation function, so I modified the public APIs to reflect this change. Let me know if I'm understanding correctly.

qwang98:
> error: failed to select a version for `ark-serialize`.
>     ... required by package `ark-ec v0.4.0`
>     ... which satisfies dependency `ark-ec = "^0.4.0"` of package `ark-bn254 v0.4.0`
>     ... which satisfies dependency `ark-bn254 = "^0.4.0"` of package `benchmark v0.1.0 (https://github.com/qwang98/plonkish.git?branch=main#e0125fe8)`
>     ... which satisfies git dependency `hyperplonk_benchmark` of package `chiquito v0.1.2023110800 (/workspaces/chiquito)`
> versions that meet the requirements `^0.4.0` are: 0.4.2, 0.4.1, 0.4.0
> 
> all possible versions conflict with previously selected packages.
> 
>   previously selected package `ark-serialize v0.4.2-alpha.1`
>     ... which satisfies dependency `ark-serialize = "^0.4.2-alpha.1"` of package `ark-ff v0.4.2`
>     ... which satisfies dependency `ark-ff = "^0.4.0"` of package `ark-ec v0.4.0`
>     ... which satisfies dependency `ark-ec = "^0.4.0"` of package `ark-bn254 v0.4.0`
>     ... which satisfies dependency `ark-bn254 = "^0.4.0"` of package `benchmark v0.1.0 (https://github.com/qwang98/plonkish.git?branch=main#e0125fe8)`
>     ... which satisfies git dependency `hyperplonk_benchmark` of package `chiquito v0.1.2023110800 (/workspaces/chiquito)`
> 
> failed to select a version for `ark-serialize` which could resolve this conflict

I think I got a similar error previously, which was fixed after I updated the rust tool chain to the latest nightly. Could you try the same to see if it's fixed?

```
(base) ➜  chiquito git:(steve/hyperplonk-backend-v2) rustc --version
rustc 1.77.0-nightly (e51e98dde 2023-12-31)
```

leolara:
So rust toolchain is defined here https://github.com/privacy-scaling-explorations/chiquito/blob/main/rust-toolchain ; And we want to keep it in sync with https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/rust-toolchain so you can advance it to that value and see if it works

qwang98:
> So rust toolchain is defined here https://github.com/privacy-scaling-explorations/chiquito/blob/main/rust-toolchain ; And we want to keep it in sync with https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/rust-toolchain so you can advance it to that value and see if it works

Yep. I just updated rust version to the one specified in zk-evm (the one we want to bring to eventually) and it works. :)

leolara:
@qwang98 you need to update it in this PR so the tests can pass

leolara:
@qwang98 I mean in rust-toolchain file, perhaps also in .github files if the toolchain is also there

leolara:
@qwang98 it seems that it is more difficult that it seems,. will require to update the dependencies of plaf to use the same halo2 that current zkEVM is using

qwang98:
@leolara I think it should be all fixed now!

Note that Clippy suggests a change for poly.rs, which I adopted but keeps bringing in more errors, so I ended up not modifying this file. This file is also not a part of this PR, so I figured maybe it's been decided to leave this Clippy error in the past.

Otherwise should be all good.

rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2023-04-24
+nightly-2024-02-13
```
leolara, line 1: why not nightly-2024-02-14 ? that is the current in zkEVM?

leolara (approved):

PR #157, opened 2023-10-20
WIP: Issue 75 add unit tests to DSL

nullbitx8:
This is a work in progress for https://github.com/privacy-scaling-explorations/chiquito/issues/75

I am planning to add code coverage as well as more tests.

Issue #105, opened 2023-08-23
Add unit tests to SuperCircuitContext


leolara:
Here: https://github.com/privacy-scaling-explorations/chiquito/blob/main/src/frontend/dsl/sc.rs

0xvon:
Hi, can I work on it?

leolara:
Assigned to you @0xvon

PR #189, opened 2024-01-14
Feature/improve unit tests coverage

rutefig:
Added more unit tests to the following modules:
- poly
- ast/query
- dsl
- dsl/cb
- compiler/step_selector
- compiler
- super_circuit

Covers some of these issues:
#157 #102 #105


rutefig:
@leolara and @qwang98 can you review it please?

src/frontend/dsl/mod.rs
```diff
@@ -154,6 +154,8 @@ impl<F, TraceArgs> CircuitContext<F, TraceArgs> {
         self.circuit.last_step = Some(step_type.into().uuid());
     }
 
+    /// Enforce the number of steps by adding a constraint to the circuit. Takes a `usize`
```
leolara, line 157: @rutefig step **instances** to distiguish form step **types**

src/plonkish/compiler/mod.rs
```diff
@@ -556,3 +556,99 @@ fn add_halo2_columns<F, TraceArgs>(unit: &mut CompilationUnit<F>, ast: &astCircu
...
+        assert_eq!(unit.ast_id, mock_ast_circuit.id);
+
+        assert!(assignment_generator.is_none());
+    }
+
+    #[test]
+    fn test_compile_phase2() {
+        let mock_ast_circuit = astCircuit::<Fr, Any>::default();
+
+        let mut unit = CompilationUnit::from(&mock_ast_circuit);
```
leolara, line 628: I don't think it makes sense to compile phase2 wihtout doing phase1 before. Perhaps we should have an error if this happens.

src/plonkish/ir/sc.rs
```diff
@@ -12,6 +12,23 @@ pub struct SuperCircuit<F, MappingArgs> {
     mapping: MappingGenerator<F, MappingArgs>,
 }
 
+impl<F: Debug, MappingArgs: Debug> Debug for SuperCircuit<F, MappingArgs> {
```
leolara, line 15: @rutefig are these implementations different from the derive ones?

leolara (commented):
check those comments

src/plonkish/compiler/mod.rs
rutefig: Agreed, so should we create an issue to fix this by tracking the phase of the compilation and ensuring that is always sequential? It should be quick but since it goes out of the unit tests scope maybe it's better to separate concerns

src/plonkish/ir/sc.rs
rutefig: Only for the `MappingArgs` - removed the one for `SuperCircuit` and added derived one

src/plonkish/compiler/mod.rs
leolara: Yes we could add a flag to compilation unit for this

rutefig: Created a new issue for this: https://github.com/privacy-scaling-explorations/chiquito/issues/192
Can I have that assigned and can we merge this PR? @leolara

src/frontend/dsl/mod.rs
```diff
@@ -562,14 +589,22 @@ mod tests {
...
+    #[test]
+    // #[should_panic(expected = "Signal not found")] // or however your system should behave
+    fn test_expose_non_existing_signal() {
+        let mut context = setup_circuit_context::<i32, i32>();
+        let non_existing_signal =
+            Queriable::Forward(ForwardSignal::new_with_phase(0, "".to_owned()), false); // Create a signal not added to the circuit
+        context.expose(non_existing_signal, ExposeOffset::First);
+
+        // TODO: Should we let the user know that they are exposing a non-existing signal? And let
+        // the circuit have an exposed signal that does not exist?
```
leolara, line 601: @rutefig add check that the signal is in the circuit, or create task for it

src/plonkish/compiler/mod.rs
leolara: @rutefig Still, the result we need to test is after compile_phase1, then compile_phase2, please dont leave the test that does not execture phase1 before phase2

src/plonkish/ir/sc.rs
```diff
@@ -7,11 +7,20 @@ use super::{
...
 
+#[derive(Debug)]
 pub struct SuperCircuit<F, MappingArgs> {
     sub_circuits: Vec<Circuit<F>>,
     mapping: MappingGenerator<F, MappingArgs>,
 }
 
+impl<F: Debug, MappingArgs: Debug> Debug for MappingGenerator<F, MappingArgs> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("MappingGenerator")
```
leolara, line 18: @rutefig why is the default Debug not good? it will not print the pointer, right?

src/plonkish/ir/sc.rs
```diff
@@ -114,3 +124,120 @@ impl<F: Field + Hash, MappingArgs> MappingGenerator<F, MappingArgs> {
...
+mod test {
+    use halo2curves::bn256::Fr;
+
+    use super::*;
+
+    #[test]
+    fn test_default() {
+        let super_circuit: SuperCircuit<Fr, ()> = Default::default();
+
+        assert_eq!(
```
leolara, line 138: @rutefig as the typed is ensure by rust, you are just checking that both vectors are empty in a very convoluted way, just check that super_circuit.sub_circuits.len == 0

src/plonkish/ir/sc.rs
```diff
@@ -114,3 +124,120 @@ impl<F: Field + Hash, MappingArgs> MappingGenerator<F, MappingArgs> {
...
+    #[test]
+    fn test_set_mapping() {
+        let mut super_circuit: SuperCircuit<Fr, ()> = Default::default();
+
+        super_circuit.set_mapping(|ctx, args| {
+            // Your mapping logic here
+            ctx.map(&AssignmentGenerator::<Fr, ()>::default(), args)
+        });
+
+        assert_eq!(
```
leolara, line 184: @rutefig what does this test?

src/plonkish/ir/sc.rs
```diff
@@ -114,3 +124,120 @@ impl<F: Field + Hash, MappingArgs> MappingGenerator<F, MappingArgs> {
...
+            format!("{:#?}", ctx.assignments),
+            format!("{:#?}", SuperAssignments::<Fr>::default())
+        );
+    }
+
+    #[test]
+    fn test_mapping_context_map() {
+        let mut ctx = MappingContext::<Fr>::default();
```
leolara, line 214: @rutefig there are many tests in this file that the defaults are defaults, that I think it is not interesting. What about actually putting some data and doing some tests about the data processing

src/poly/mod.rs
```diff
@@ -322,4 +322,124 @@ mod test {
...
+    #[test]
+    fn neg_of_queries_and_constants() {
+        use super::Expr::*;
+
+        let a: &str = "a";
+        let b: &str = "b";
+        let c: Fr = Fr::from(3);
+        let neg_expr: Expr<Fr, &str> = Query(a) * Query(b) * Const(c) - Const(c);
+
+        assert!(
```
leolara, line 397: @rutefig you should check that the rhs is Neg too

src/sbpir/query.rs
```diff
@@ -211,4 +211,243 @@ mod tests {
         let expr5: Expr<Fr, Queriable<Fr>> = Expr::Pow(Box::new(Expr::Const(a)), 2);
         assert_eq!(format!("{:?}", expr5), "(0xa)^2");
     }
+
+    #[test]
+    fn test_add_fmt() {
+        let a: Fr = 10.into();
```
leolara, line 217: @rutefig this test does not test something specific about this file?

leolara: To test the operator you need to test with lhs and another with with rhs as queriables, otherwise you are using the expression operator

src/plonkish/ir/sc.rs
rutefig: Yes, it does not print the pointer

src/plonkish/ir/sc.rs
leolara: You marked as resolved but the change is not done

leolara:
@rutefig it has conflicts

src/sbpir/query.rs
rutefig: You're right, removed as it didn't make sense

src/plonkish/ir/sc.rs
rutefig: But if we don't print the pointer we still need to implement Debug right? we cannot use the default one because the pointer does not implements Debug

leolara: I think you can use the default one, it just ignore the pointer

src/plonkish/ir/sc.rs
leolara: @rutefig can you do this?

src/poly/mod.rs
```diff
@@ -322,4 +322,124 @@ mod test {
...
+    fn neg_of_queries_and_constants() {
+        use super::Expr::*;
+
+        let a: &str = "a";
+        let b: &str = "b";
+        let c: Fr = Fr::from(3);
+        let neg_expr: Expr<Fr, &str> = Neg(Box::new(Query(a) * Query(b) * Const(c) - Const(c)));
+
+        assert!(
+            matches!(neg_expr, Expr::Neg(_)),
```
leolara, line 398: @rutefig of course neg_expr is Neg, because you are just making it so a few lines before. Perhaps you can cehck that lhs - rhs, produces Sum(lhs, Neg(rhs)).

src/poly/mod.rs
```diff
@@ -322,4 +322,124 @@ mod test {
...
+    fn pow_of_queries() {
+        use super::Expr::*;
+
+        let a: &str = "a";
+        let pow_expr: Expr<Fr, &str> = Pow(Box::new(Query("a")), 2);
+        let mut assignments: VarAssignments<Fr, &str> = VarAssignments::default();
+        assignments.insert(a, Fr::from(2));
+
+        assert!(
+            matches!(pow_expr, Expr::Pow(_, _)),
```
leolara, line 429: @rutefig you made it equal to Pow a few lines before. 

Also you are creating this assignments but not using it

src/plonkish/ir/sc.rs
rutefig: It complaints if I try to use the default and it doesn't allow to use it

src/frontend/dsl/sc.rs
```diff
@@ -120,3 +121,52 @@ where
...
+        );
+        assert_eq!(ctx.sub_circuit_phase1.len(), 0);
+        assert_eq!(
+            format!("{:#?}", ctx.tables),
+            format!("{:#?}", LookupTableRegistry::<Fr>::default())
+        );
+    }
+
+    #[test]
+    fn test_super_circuit_context_sub_circuit() {
```
leolara, line 155: @rutefig you should add tests that test all the methods, like sub_circuit_with_ast, mapping and compile. They executed with some data that makes sense and then test that the result make sense. If some methods only make sense after executing others, execute them in that order

leolara (approved):

leolara:
@rutefig it is approved, but you need to fix clippy and rustfmt before merging

PR #190, opened 2024-01-16
Experimental parser

leolara:


leolara:
@ed255 this does not merge to main, but to chiquito-2024 so it is ok that it is incomplete

src/parser/ast/expression.rs
```diff
@@ -0,0 +1,215 @@
...
+impl BinaryOperator {
+    pub fn is_logic(&self) -> bool {
+        use BinaryOperator::*;
+        match self {
+            Eq => true,
+            NEq => true,
+            And => true,
+            Or => true,
+            Sum => false,
+            _ => false,
```
ed255, line 26:
```suggestion
            _ => false,
```

src/parser/ast/expression.rs
```diff
@@ -0,0 +1,215 @@
...
+        use BinaryOperator::*;
+        match value.as_str() {
+            "+" => Sum,
+            "*" => Mul,
+            "^" => Pow,
+            "==" => Eq,
+            "!=" => NEq,
+            "&&" => Or,
+            "||" => And,
+            &_ => unimplemented!(),
```
ed255, line 61: unimplemented? or unreachable?  Will this be an Error in the future?

src/parser/ast/expression.rs
```diff
@@ -0,0 +1,215 @@
...
+        match value.as_str() {
+            "!" => Not,
+            "-" => Neg,
+            &_ => unimplemented!(),
+        }
+    }
+}
+
+#[derive(Clone)]
+pub enum Expression<F, V> {
```
ed255, line 108: How about this?
```rust
pub enum Expression<F, V> {
    BinOp {
        op: BinaryOperator,
        lhs: Box<Node<F, V>>,
        rhs: Box<Node<F, V>>,
    },
    // [...]
}

pub struct Node<F, V>{
  dsym: DebugSymbRef,
  e: Expression<F, V>,
}
```

This way you don't need to add dsym in each possible enum case.

src/parser/ast/expression.rs
```diff
@@ -0,0 +1,215 @@
...
+        rhs: Box<Expression<F, V>>,
+    },
+
+    UnaryOp {
+        dsym: DebugSymRef,
+        op: UnaryOperator,
+        sub: Box<Expression<F, V>>,
+    },
+
+    Select(
```
ed255, line 123: what about field names for this?  Example:
- condition
- case_true
- case_false

src/parser/ast/expression.rs
```diff
@@ -0,0 +1,215 @@
...
+    False(DebugSymRef),
+}
+
+impl<F, V> Expression<F, V> {
+    pub fn is_arith(&self) -> bool {
+        use Expression::*;
+        match self {
+            BinOp { op, .. } => op.is_arith(),
+            UnaryOp { op, .. } => op.is_arith(),
+            Select(_, _, _, _) => true,
```
ed255, line 143: what about this:  select(x == y) {z == w} else {false}.  This is a select that returns a logic expression.
I think the select should have the same type in each case, and evaluate to that type.

src/parser/ast/expression.rs
```diff
@@ -0,0 +1,215 @@
...
+            Query(_, _) => true,
```
ed255, line 144: What about boolean signals?

src/parser/ast/expression.rs
```diff
@@ -0,0 +1,215 @@
...
+            Const(_, _) => true,
```
ed255, line 145: What about boolean constants?

src/parser/ast/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+
+pub mod expression;
+pub mod statement;
+pub mod tl;
+
+#[derive(Clone, Debug)]
+pub struct DebugSymRef {
+    pub start: usize,
+    pub end: usize,
+}
```
ed255, line 11: Is this a TODO?  At this stage I think it's OK to have unfinished parts, but I will request that at least there are some comments with TODO, with a small explanation of what's the intention of the types.

Please, add as many TODOs as necessary to this PR!

src/parser/ast/statement.rs
```diff
@@ -0,0 +1,61 @@
+use std::fmt::Debug;
+
+use super::{expression::Expression, DebugSymRef};
+
+#[derive(Clone)]
+pub enum Statement<F, V> {
+    AssertEq(DebugSymRef, Expression<F, V>, Expression<F, V>),
+    AssertNEq(DebugSymRef, Expression<F, V>, Expression<F, V>),
```
ed255, line 8: why not assert of logical expression?

This way we can do:
```
assert is_foo and is_bar
```
or
```
assert is_foo or is_bar
```
Then an eq is an expression that returns an logical

src/parser/ast/statement.rs
```diff
@@ -0,0 +1,61 @@
...
+        DebugSymRef,
+        Box<Expression<F, V>>,
+        Vec<Statement<F, V>>,
+        Vec<Statement<F, V>>,
+    ),
+
+    SignalDecl(DebugSymRef, String),
+    ForwardSignalDecl(DebugSymRef, String),
+    InternalSignalDecl(DebugSymRef, String),
+    VarDecl(DebugSymRef, String),
```
ed255, line 23: Can I also declare constants?

src/parser/ast/expression.rs
ed255, line 108: This Expression type works as a binary tree with binary operations.  So if you do `a + b + c + d` you get `a + (b + (c + d))`.
In Chiquito the expression type uses Vec for sums and muls.  Did you think about this?

ed255 (commented):
Initial review.  One request I have is to write the grammar in a text file in a concise form, which will make it easier to discuss and iterate without requiring a full implementation.  Something like a BNF grammar, or something like this https://craftinginterpreters.com/appendix-i.html

Also I believe finding a nice syntax and grammar for this may take some iterations, so I wonder what's the best way to stay flexible during the exploration, so that changing parts of the grammar isn't too painful.

src/parser/ast/expression.rs
rutefig: Sorry for the intromission here, but actually I think unimplemented can make sense because it means that anything else besides what was defined previously is not implemented, right?

leolara: Yes, it is a way to throw an error. This is called by the parser and as the user does not have a way to include new operators calling this with a different symbol would be a bug, not an error that the user could trigger, so I would not implement an actual error reporting system for this.

src/parser/ast/expression.rs
leolara: But expression is not made or nodes, so we would only get the DebugSymRef in the root, the idea is that we will have the DebugSymRef in each node of the AST.

src/parser/ast/expression.rs
leolara: It is a way of doing it, but I thought this is simpler. We could do a rule and do this every node?

src/parser/ast/expression.rs
leolara: Select is ? : that returns arithmetic value, as different from if then else that is an assertion. cond of select should be logic, when_true and when_false should be arithmetic.

src/parser/ast/expression.rs
leolara: There are True and False as their own AST nodes

src/parser/ast/mod.rs
leolara: DebugSymRef will grow with more data as we add more information to errors and debug information, currently it is a placeholder, that such info should be in each node of the AST.

Everything is a big TODO, but I will add a TODO for this.

src/parser/ast/statement.rs
leolara: We will be able to, this is incomplete.

src/parser/ast/expression.rs
leolara: Yes, this AST reflects the grammar that is based on binary operations. We will have to implement a flattening of the chain of binaries for optimal compilation. I was think perhaps substituting lhs, rhs by an array, the parser will only produce binary, but then we can flattening in a post-process.

leolara:
> Initial review. One request I have is to write the grammar in a text file in a concise form, which will make it easier to discuss and iterate without requiring a full implementation. Something like a BNF grammar, or something like this https://craftinginterpreters.com/appendix-i.html
> 
> Also I believe finding a nice syntax and grammar for this may take some iterations, so I wonder what's the best way to stay flexible during the exploration, so that changing parts of the grammar isn't too painful.

So I think there are several posibilities:

 + Create a file with BNF. I think the problem with this is to keep syncing the files
  + Add BNF as comments to the lalrpop file. Circom doesn't seem to need this: https://github.com/iden3/circom/blob/master/parser/src/lang.lalrpop
  + Add unit tests with examples of the currently working syntaxt. I think that could be a way to show the syntaxt, test and don't add an extra layer.
  
  It is completely true that an LR1 grammar is more difficult to read that BNF as it is more technical and limited. The parser only sees one token ahead.

src/parser/ast/expression.rs
leolara: Yes, it should be unreachable

src/parser/ast/expression.rs
ed255: > But expression is not made or nodes, so we would only get the DebugSymRef in the root, the idea is that we will have the DebugSymRef in each node of the AST.

I don't understand what you're saying.  Why do you claim that expression is not made of nodes?  In the snippet I wrote, Expressions are made of nodes, so you get a DebugSYmRef in each node.

src/parser/ast/expression.rs
ed255: I know that select is the `x ? y : z` operation from other languages.  I'm asking: why can't the output of that expression be logical?

src/parser/ast/expression.rs
leolara: Ahh, I didn't see the sub expressions are node. I don't know, it will create these two structs that are always including each other, in one way it is more separation of concerns but in other will make the code more difficult perhaps?

src/parser/ast/expression.rs
leolara: Done

src/parser/ast/expression.rs
leolara: Done

src/parser/ast/mod.rs
leolara: TODO added

src/parser/ast/expression.rs
leolara: For that we have the if/then/else that is a different statement, to allow this to be logical would complicate things.

src/parser/ast/expression.rs
leolara: And alternative would be to give an ID to each node, and have symbol table that references to the DebugSymRef

src/parser/ast/statement.rs
leolara: Added generic assert statement

src/parser/ast/expression.rs
leolara: I already changed it so the expressions contain an array, and i already implemeted in the compiler (next PR) to flatten the binary tree before compiling.

leolara: Let me clarify, the target Expression contains and array instead of bianry. The compiler flatten the birary tree and produces expressions with arrays as sub expressions

ed255: > Ahh, I didn't see the sub expressions are node. I don't know, it will create these two structs that are always including each other, in one way it is more separation of concerns but in other will make the code more difficult perhaps?

I have the intuition that my suggestion may lead to simpler code; but I just wanted to give you the suggestion (not a request).  Feel free to make the choice that seems better to you for the implementation :)

> And alternative would be to give an ID to each node, and have symbol table that references to the DebugSymRef

This would just replace `DebugSymRef` by an `ID` right?  So it doesn't seem an alternative to the suggestion, as it applies in both your original code and my suggestion.

leolara: Yes, but ID is more flexible because it can lead to different tables with different information. So it is more decoupled

src/parser/ast/expression.rs
leolara: I have made the select to be able to have a logic result

leolara:
I think it is ready for another pass @ed255

ed255 (approved):
LGTM!
Nevertheless I will still ask that you write the grammar into BNF/EBNF format!

Issue #192, opened 2024-01-23
Track compilation phases and not allow phase2 without phase1


rutefig:
I don't think it makes sense to compile phase2 wihtout doing phase1 before. Perhaps we should have an error if this happens.

_Originally posted by @leolara in https://github.com/privacy-scaling-explorations/chiquito/pull/189#discussion_r1452214219_

PR #195, opened 2024-01-30
Track compilation phase on CompilationUnit

rutefig:
**Issue:** #192 
It adds a field on the CompilationUnit struct to store the phase of compilation, and panics if we try to run `compile_phase2` without phase1 being done before.

### DIFF
--- a/src/plonkish/compiler/mod.rs
+++ b/src/plonkish/compiler/mod.rs
@@ -95,10 +95,16 @@ pub fn compile_phase1<
         )
     });
 
+    unit.compilation_phase = 1;
+
     (unit, assignment)
 }
 
 pub fn compile_phase2<F: Field + Clone>(unit: &mut CompilationUnit<F>) {
+    if unit.compilation_phase != 1 {
+        panic!("Compilation phase 2 can only be done after compilation phase 1");
+    }
+
     for step in unit.step_types.clone().values() {
         compile_step(unit, step);
     }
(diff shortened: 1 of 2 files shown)


src/plonkish/compiler/unit.rs
```diff
@@ -48,6 +48,8 @@ pub struct CompilationUnit<F> {
 
     pub other_sub_circuits: Rc<Vec<CompilationUnit<F>>>,
     pub other_columns: Rc<Vec<Column>>,
+
+    pub compilation_phase: usize,
```
leolara, line 52: @rutefig we use usize for something that indexes an array or is the size of something, perhaps it should be u32. But I actually I would do phase1_complete:bool. Because we are not thinking in having more phases.

src/plonkish/compiler/mod.rs
```diff
@@ -95,10 +95,16 @@ pub fn compile_phase1<
         )
     });
 
+    unit.compilation_phase = 1;
+
     (unit, assignment)
 }
 
 pub fn compile_phase2<F: Field + Clone>(unit: &mut CompilationUnit<F>) {
+    if unit.compilation_phase != 1 {
```
leolara, line 104: @rutefig we could also panic if you try to do phase1 twice

src/plonkish/compiler/unit.rs
rutefig: I agree we can change to u32 for convention, but I think it's better to have the phase tracked anyways because it can give more flexibility in the future. Even if we don't have any more phases, we can track at some point that phase 2 is also already completed

src/plonkish/compiler/mod.rs
rutefig: checking the implications of that

leolara (approved):

PR #198, opened 2024-02-04
Rutefig/194 add check that the signal is in the circuit

rutefig:
Adds a check if the signal that is trying to be exposed is already in the circuit or not, this way the user will always know if is trying to expose a non existing signal.


leolara (approved):

leolara:
Can you merge?

PR #201, opened 2024-02-11
Develop blake2f circuits

10to4:
Add blake2f circuit example.

References

* [rfc7693](https://datatracker.ietf.org/doc/html/rfc7693#section-3.2)
* [eip-152](https://github.com/ethereum/EIPs/blob/6572e92dccb2a581c0082befb953050f75d0ece5/EIPS/eip-152.md)


examples/blake2f.rs
```diff
@@ -0,0 +1,1371 @@
...
+    [7, 9, 3, 1, 13, 12, 11, 14, 2, 6, 5, 10, 4, 0, 15, 8],
+    [9, 0, 5, 7, 2, 4, 10, 15, 14, 1, 11, 12, 6, 8, 3, 13],
+    [2, 12, 6, 10, 0, 11, 8, 3, 4, 13, 7, 5, 15, 14, 1, 9],
+    [12, 5, 1, 15, 14, 13, 4, 10, 0, 7, 6, 3, 9, 2, 8, 11],
+    [13, 11, 7, 14, 12, 1, 3, 9, 5, 0, 15, 4, 8, 6, 2, 10],
+    [6, 15, 14, 9, 11, 3, 0, 8, 12, 2, 13, 7, 1, 4, 10, 5],
+    [10, 2, 8, 4, 7, 6, 1, 5, 15, 11, 9, 14, 3, 12, 13, 0],
+];
+
+pub const XOR_VALUES: [u64; XOR_4BITS_NUMBER] = [
```
leolara, line 64: Why u64, should be u8 enought?

examples/blake2f.rs
```diff
@@ -0,0 +1,1371 @@
...
+    0, 3, 2, 13, 12, 15, 14, 9, 8, 11, 10, 6, 7, 4, 5, 2, 3, 0, 1, 14, 15, 12, 13, 10, 11, 8, 9, 7,
+    6, 5, 4, 3, 2, 1, 0, 15, 14, 13, 12, 11, 10, 9, 8, 8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2, 3, 4,
+    5, 6, 7, 9, 8, 11, 10, 13, 12, 15, 14, 1, 0, 3, 2, 5, 4, 7, 6, 10, 11, 8, 9, 14, 15, 12, 13, 2,
+    3, 0, 1, 6, 7, 4, 5, 11, 10, 9, 8, 15, 14, 13, 12, 3, 2, 1, 0, 7, 6, 5, 4, 12, 13, 14, 15, 8,
+    9, 10, 11, 4, 5, 6, 7, 0, 1, 2, 3, 13, 12, 15, 14, 9, 8, 11, 10, 5, 4, 7, 6, 1, 0, 3, 2, 14,
+    15, 12, 13, 10, 11, 8, 9, 6, 7, 4, 5, 2, 3, 0, 1, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3,
+    2, 1, 0,
+];
+
+fn split_value_4bits<F: PrimeField + Hash>(value: u128, n: usize) -> Vec<F> {
```
leolara, line 77: You can make a parameter mutable, `(mut value: u128` so it is not necessary the first line

examples/blake2f.rs
```diff
@@ -0,0 +1,1371 @@
...
+        for (i, &value) in iv_values.iter().enumerate().take(IV_LENGTH) {
+            ctx.assign(i, lookup_iv_row, F::from(i as u64));
+            ctx.assign(i, lookup_iv_value, F::from(value));
+        }
+    });
+
+    ctx.new_table(table().add(lookup_iv_row).add(lookup_iv_value))
+}
+
+fn blake2f_4bits_table<F: PrimeField + Hash>(
```
leolara, line 132: @10to4 I don't understand this table that both values are the same? Is it to see if the value is 4bits? In that case I would have a signal that is 1 always for all rows, and the other signal with `i`, Then you lookup to (1, x) always.

examples/blake2f.rs
```diff
@@ -0,0 +1,1371 @@
...
+            })
+            .collect();
+        let setup_final_split_bits_vec = final_split_bits_vec.clone();
+
+        ctx.setup(move |ctx| {
+            // check inputs: h_vec
+            for (i, split_4bits_vec) in h_split_4bits_vec.iter().enumerate() {
+                let mut h_4bits_sum_value = 0.expr() * 1;
+                for &bits in split_4bits_vec.iter().rev() {
+                    ctx.add_lookup(params.bits_table.apply(bits).apply(bits));
```
leolara, line 372: You could make a function that is called `check_4bit(ctx, a)` and does `ctx.add_lookup(params.bits_table.apply(bits).apply(bits));` that would be more readable.

examples/blake2f.rs
```diff
@@ -0,0 +1,1371 @@
...
+        .map(|i| ctx.forward(format!("h_vec[{}]", i).as_str()))
+        .collect();
+    let m_vec: Vec<Queriable<F>> = (0..M_LEN)
+        .map(|i| ctx.forward(format!("m_vec[{}]", i).as_str()))
+        .collect();
+    let round = ctx.forward("round");
+
+    let blake2f_pre_step = ctx.step_type_def("blake2f_pre_step", |ctx| {
+        let v_vec = v_vec.clone();
+        let setup_v_vec = v_vec.clone();
```
leolara, line 310: @10to4 setup_v_vec seems to be used in wg, perhaps should be called `wg_v_vec?

Cannot you do the clone inside wg, so it is more clear?

examples/blake2f.rs
10to4: It is because this parameter maybe a sum of two or three values.
For example:
```rust 
va_bit_vec.push(split_value_4bits(
    v1_vec_values[a] as u128 + v1_vec_values[b] as u128 + x as u128,
    BITS_NUMBER + 1,
));
```
So if we make the parameter mutable, we need to defined a new mutable value outside before we call the split_value_4bits function. Code readability doesn't seem to be better.

examples/blake2f.rs
10to4: Make sense. But I am not sure if all the keys are same, there will be potential problems in lookup table queries if we integrate more lookup schemes on Chiquito in the future. Because I'm not sure if all lookup table schemes allow the same key to exist.

examples/blake2f.rs
10to4: Ok, Got it.

examples/blake2f.rs
10to4: Yes, that's right!

examples/blake2f.rs
10to4: Yes, it is.

examples/blake2f.rs
leolara: What you match are tuples, so it should be possible.

examples/blake2f.rs
leolara: I don't think that is the case. u128 implements the Copy trait. So you can pass a immutatble variable to a mutable parameter, because it is always a copy in the parameter.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=341b4f4b4eb26ca6fb409f2052349ea4

examples/blake2f.rs
10to4: The reason we clone vectors is that we can't use a same vector in both the setup function and the wg function, so we may be able to do the clone in wg, because that means we need to use v_vec in both the setup function and the wg function

leolara: yes, I understand why we need to clone, but it is confusing that `setup_` is used in wg instead of setup. also if you can clone in the same level of where it is used, that is more clear.

examples/blake2f.rs
```diff
@@ -0,0 +1,1377 @@
...
+                let setup_b_bit_vec = b_bit_vec.clone();
+                let b_3bits_vec: Vec<Queriable<F>> = (0..G_ROUND / 2)
+                    .map(|i| ctx.internal(format!("b_3bits_vec[{}]", i).as_str()))
+                    .collect();
+                let setup_b_3bits_vec = b_3bits_vec.clone();
+
+                ctx.setup(move |ctx| {
+                    let s = SIGMA_VALUES[r % 10];
+
+                    // Rotation constants | (R1, R2, R3, R4)  | = | (32, 24, 16, 63)
```
leolara, line 682: @10to4 I think this comment here it is confusing, because you are not using the R* constants right away, and anyway the constants are named already. And the values shouldn't change as we are only doing black2b

examples/blake2f.rs
```diff
@@ -0,0 +1,1377 @@
...
+                    for i in 0..G_ROUND {
+                        let mut input_vec = v_vec.clone();
```
leolara, line 684: @10to4 I think the content of this loop is too long, perhaps you should extract it to a function called g_setup or something like that, that receives a borrow of the context and the signals. Perhaps rename `g_internal` to `g_wg` or `g_witness`

examples/blake2f.rs
```diff
@@ -0,0 +1,1377 @@
...
+pub const R3: usize = 16;
+pub const R4: usize = 63;
+pub const MIXING_ROUND: usize = 12;
+pub const BITS_NUMBER: usize = 16;
+pub const VALUE_4BITS: u64 = 16;
+pub const XOR_4BITS_NUMBER: usize = BITS_NUMBER * BITS_NUMBER;
+pub const V_LEN: usize = 16;
+pub const M_LEN: usize = 16;
+pub const H_LEN: usize = 8;
+pub const G_ROUND: usize = 16;
```
leolara, line 38: G_ROUNDS in plural because it is the amount of times we need to do something.

examples/blake2f.rs
```diff
@@ -0,0 +1,1377 @@
...
+};
+use halo2_proofs::{
+    dev::MockProver,
+    halo2curves::{bn256::Fr, group::ff::PrimeField},
+};
+use std::hash::Hash;
+
+pub const IV_LENGTH: usize = 8;
+pub const SIGMA_VECTOR_LENGTH: usize = 16;
+pub const R1: usize = 32;
```
leolara, line 27: Perhaps we are using usize here where we should use u32 or u64. usize is for indexing arrays, their lenght or ranges. But not for any integer.

https://www.reddit.com/r/rust/comments/4lql5l/beginner_question_should_i_use_usize_and_isize/

examples/blake2f.rs
```diff
@@ -0,0 +1,1377 @@
...
+                for (&q, &v) in q_vec.iter().zip(v_vec.iter()) {
+                    ctx.assign(q, v)
+                }
+            }
+        })
+    });
+
+    let blake2f_g_step_vec: Vec<StepTypeWGHandler<F, _, _>> = (0..MIXING_ROUND)
+        .map(|r| {
+            ctx.step_type_def("blake2f_g_step", |ctx| {
```
leolara, line 582: @10to4 perhaps use format! so the name of the step type contains `r` value?

examples/blake2f.rs
10to4: That's right, I'll remove it.

examples/blake2f.rs
10to4: Whether we can't contain `r` value while using the function below?

```rust
impl From<&'static str> for StepTypeDefInput {
    fn from(s: &'static str) -> Self {
        StepTypeDefInput::String(s)
    }
}
```

It seems we need to add a new function `fn from(s: &str) -> Self ` or `fn from(s: String) -> Self` function for StepTypeDefInput.

leolara: I don't understand exactly but you can add the `From for StepTypeDefInput` for the type that format! returns.

examples/blake2f.rs
```diff
@@ -0,0 +1,1427 @@
...
+            // check inputs: m_vec
+            for (i, split_4bits_vec) in m_split_4bits_vec.iter().enumerate() {
+                let mut m_4bits_sum_value = 0.expr() * 1;
+                for &bits in split_4bits_vec.iter().rev() {
+                    params.check_4bit(ctx, bits);
+                    m_4bits_sum_value = m_4bits_sum_value * VALUE_4BITS + bits;
+                }
+                ctx.constr(eq(m_4bits_sum_value, m_vec[i]))
+            }
+            // check inputs: t0,t1
```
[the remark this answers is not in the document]
10to4: t is the merge of t0 and t1, the low 64 bits are t0 and the high 64 bits are t1. 
[rfc7693](https://datatracker.ietf.org/doc/html/rfc7693#section-3.2)
> 2w-bit offset counter "t"

> v[12] := v[12] ^ (t mod 2**w) // Low word of the offset.
 v[13] := v[13] ^ (t >> w) // High word.

At the docment [eip-152](https://github.com/ethereum/EIPs/blob/6572e92dccb2a581c0082befb953050f75d0ece5/EIPS/eip-152.md), it is defined as 2 values.
> t_0, t_1 - offset counters - 2 unsigned 64-bit little-endian words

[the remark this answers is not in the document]
leolara: Very smart

examples/blake2f.rs
```diff
@@ -0,0 +1,1427 @@
...
+                .zip(v_vec_values[V_LEN as usize / 2..V_LEN as usize].iter())
+                .map(|(&v1, &v2)| split_xor_value(v1, v2))
+                .collect(),
+            final_split_bit_vec: output_vec_values
+                .iter()
+                .map(|&output| split_value_4bits(output as u128, BITS_NUMBER))
+                .collect(),
+        };
+        ctx.add(&blake2f_final_step, final_inputs);
+        // ba80a53f981c4d0d, 6a2797b69f12f6e9, 4c212f14685ac4b7, 4b12bb6fdbffa2d1
```
leolara, line 1347: @10to4 have you checked this result with other implementation, even better with the EVM implementation?

examples/blake2f.rs
```diff
@@ -0,0 +1,1427 @@
...
+fn g_setup<F: PrimeField + Hash>(
+    ctx: &mut StepTypeSetupContext<'_, F>,
+    params: CircuitParams,
+    q_params: GStepParams<F>,
+    (a, b, c, d): (usize, usize, usize, usize),
+    (move1, move2): (u64, u64),
+    s: usize,
+    flag: bool,
+) {
+    // check v[a] = (v[a] + v[b] + x) mod 2^64
```
leolara, line 365: @10to4 it seems that here you are doing:

```
 v[a] := (v[a] + v[b] + x) mod 2**w
v[d] := (v[d] ^ v[a]) >>> R1
v[c] := (v[c] + v[d])     mod 2**w
 v[b] := (v[b] ^ v[c]) >>> R2
 ```
 
But not:

```
v[a] := (v[a] + v[b] + y) mod 2**w
v[d] := (v[d] ^ v[a]) >>> R3
 v[c] := (v[c] + v[d])     mod 2**w
 v[b] := (v[b] ^ v[c]) >>> R4
 ```
 
 Or am I missing something?

10to4: This two parts are similar. In order to reduce repetitive code, both the parts are implemented by `g_step` function with different parameters. In the spec, the `function G` is called 8 times, so we need to called the `g_step` function 16 times.

examples/blake2f.rs
10to4: Yes, the input example comes from the document [eip-152](https://github.com/ethereum/EIPs/blob/6572e92dccb2a581c0082befb953050f75d0ece5/EIPS/eip-152.md#example-usage-in-solidity).

leolara:
@10to4 I think it would make it much more readable if when you define a new array of signals you put a comment with the formula it represents. For example:

```
// v_xor_b_bit_vec[i] === 4bit split of v[i] ^ v[i+8]
let v_xor_b_bit_vec: Vec<Vec<Queriable<F>>>
```

That way it would be easier to follow the logic.

leolara:
Also with this function, i think you would eliminate a lot of code repetition and make it more readable:

```
fn split_4bit_signals<F: PrimeField + Hash>(
    ctx: &mut StepTypeSetupContext<F>,
    params: &CircuitParams,
    input: &[Queriable<F>],
    output: &[Vec<Queriable<F>>],
) {
    for (i, split_vec) in output.iter().enumerate() {
        let mut sum_value = 0.expr() * 1;

        for &bits in split_vec.iter().rev() {
            params.check_4bit(ctx, bits);
            sum_value = sum_value * VALUE_4BITS + bits;
        }
        ctx.constr(eq(sum_value, input[i]))
    }
}
```

src/frontend/dsl/mod.rs
```diff
@@ -201,6 +201,12 @@ impl From<&'static str> for StepTypeDefInput {
     }
 }
 
+impl From<String> for StepTypeDefInput {
+    fn from(s: String) -> Self {
+        StepTypeDefInput::String(Box::leak(s.into_boxed_str())) // 将 String 转换为 &'static str
```
leolara, line 206: @10to4 what is that in Chinese?

examples/blake2f.rs
```diff
@@ -0,0 +1,1459 @@
...
+
+fn main() {
+    let super_circuit = blake2f_super_circuit::<Fr>();
+    let compiled = chiquitoSuperCircuit2Halo2(&super_circuit);
+
+    let values = InputValues {
+        round: 12,
+        // h[0] = hex"48c9bdf267e6096a 3ba7ca8485ae67bb 2bf894fe72f36e3c f1361d5f3af54fa5";
+        // h[1] = hex"d182e6ad7f520e51 1f6c3e2b8c68059b 6bbd41fbabd9831f 79217e1319cde05b";
+        h_vec: [
```
leolara, line 1427: @10to4 the values in h_vec does not seem to match the comment and the EIP, why is that?

leolara: Is it because of the endianness?

examples/blake2f.rs
```diff
@@ -0,0 +1,1459 @@
...
+    poly::ToExpr,
+    sbpir::query::Queriable,
+};
+use halo2_proofs::{
+    dev::MockProver,
+    halo2curves::{bn256::Fr, group::ff::PrimeField},
+};
+use std::hash::Hash;
+
+pub const IV_LENGTH: u64 = 8;
```
leolara, line 25: @10to4 IV_LENGTH could be of type usize for example. If you use it for the length or indexing an array then usize is good. But it does not matter, you can leave it like that.

examples/blake2f.rs
```diff
@@ -0,0 +1,1459 @@
...
+pub const SIGMA_VECTOR_LENGTH: u64 = 16;
+pub const SIGMA_VECTOR_NUMBER: u64 = 10;
+pub const R1: u64 = 32;
+pub const R2: u64 = 24;
+pub const R3: u64 = 16;
+pub const R4: u64 = 63;
+pub const MIXING_ROUND: u64 = 12;
```
leolara, line 32: @10to4 if it is then number of roundS, it should be plural, MIXING_ROUNDS

examples/blake2f.rs
```diff
@@ -0,0 +1,1459 @@
...
+pub const BITS_NUMBER: u64 = 16;
```
leolara, line 33: @10to4 any number os a number. Perhaps if this is the bit width of something, BITS_WIDTH is more clear.

leolara: It is something else it seems. Like the amount of values possible for 4bits. I am not sure what name would be more clear for this

leolara: I need to think about the best name for this contant, and also if it makes sense to separate it from VALUE_4BITS

examples/blake2f.rs
```diff
@@ -0,0 +1,1459 @@
...
+    v_xor_b_bit_vec.push(split_value_4bits(
+        (v1_vec_values[b] ^ v2_vec_values[c]) as u128,
+        BITS_NUMBER,
+    ));
+    let bits = (v1_vec_values[b] ^ v2_vec_values[c]) / 2u64.pow(60);
+    b_bit_vec.push(F::from(bits / 8));
+    b_3bits_vec.push(F::from(bits % 8))
+}
+
+fn g_setup<F: PrimeField + Hash>(
```
leolara, line 356: @10to4 we need to explain in a comment that one call to g_setup just constraint half of what the G function in the spec does.

examples/blake2f.rs
```diff
@@ -0,0 +1,1459 @@
...
+                    })
+                    .collect();
+                let wg_v_xor_b_bit_vec = v_xor_b_bit_vec.clone();
+
+                // b_bit_vec[i] * 8 + b_3bits_vec[i] = v_xor_b_bit_vec[i * 2 + 1][0]
+                let b_bit_vec: Vec<Queriable<F>> = (0..G_ROUNDS / 2)
+                    .map(|i| ctx.internal(format!("b_bit_vec[{}]", i).as_str()))
+                    .collect();
+                let wg_b_bit_vec = b_bit_vec.clone();
+                let b_3bits_vec: Vec<Queriable<F>> = (0..G_ROUNDS / 2)
```
leolara, line 818: Comment anout the formula for b_3bits_vec

examples/blake2f.rs
10to4: Yes, I can continue to change the output order.

examples/blake2f.rs
```diff
@@ -0,0 +1,1459 @@
...
+            0x510e527fade682d1,
+            0x9b05688c2b3e6c1f,
+            0x1f83d9abfb41bd6b,
+            0x5be0cd19137e2179,
+        ], // 8 * 64bits
+        // m[0] = hex"6162630000000000000000000000000000000000000000000000000000000000";
+        // m[1] = hex"0000000000000000000000000000000000000000000000000000000000000000";
+        // m[2] = hex"0000000000000000000000000000000000000000000000000000000000000000";
+        // m[3] = hex"0000000000000000000000000000000000000000000000000000000000000000";
+        m_vec: [0x636261, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], // 16 * 64bits
```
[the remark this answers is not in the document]
10to4: We use 2 hex to represent a byte, at m[0], 0x61 is the first byte, 0x62 is the second byte, 0x63 is the third byte, the rest are 0. So when we combine 16 bytes to one value, it will be represent to 0x636261. Because the first byte is in the lowest bit.

examples/blake2f.rs
leolara: Ok

examples/blake2f.rs
10to4: I want to use this constant to represent the number of 4-bit values that are separated from a original value.

examples/blake2f.rs
10to4: ok

leolara:
@10to4 What do you think if the constants VALUE_4BITS and BITS_NUMBER are the same and are called BASE_4BITS ?

leolara (approved):
The PR is approved and can be merged. It only needs "the constants VALUE_4BITS and BITS_NUMBER are the same and are called BASE_4BITS".

However I would recommend some DRY before merging, although it is not required.

As an example:
```
let h_vec_values = values.h_vec.to_vec();
        let h_split_4bits_vec = h_vec_values
            .iter()
            .map(|&value| {
                let mut value = value;
                (0..BITS_NUMBER)
                    .map(|_| {
                        let v = value % VALUE_4BITS;
                        value >>= 4;
                        F::from(v)
                    })
                    .collect()
            })
            .collect();

        let m_vec_values = values.m_vec.to_vec();
        let m_split_4bits_vec = m_vec_values
            .iter()
            .map(|&value| {
                let mut value = value;
                (0..BITS_NUMBER)
                    .map(|_| {
                        let v = value % VALUE_4BITS;
                        value >>= 4;
                        F::from(v)
                    })
                    .collect()
            })
            .collect();

        let mut iv_vec_values = IV_VALUES.to_vec();
        let iv_split_4bits_vec: Vec<Vec<F>> = iv_vec_values[4..7]
            .iter()
            .map(|&value| {
                let mut value = value;
                (0..BITS_NUMBER)
                    .map(|_| {
                        let v = value % VALUE_4BITS;
                        value >>= 4;
                        F::from(v)
                    })
                    .collect()
            })
            .collect();
```
Could be make much more clear with a function, and there are other cases I think.

10to4:
> @10to4 What do you think if the constants VALUE_4BITS and BITS_NUMBER are the same and are called BASE_4BITS ?

@leolara Even though the value of VALUE_4BITS and BITS_NUMBER are the same, the meanings are different. VALUE_4BITS represents the max value of 4bits, BITS_NUMBER means the number of 4bits values we can split from a 64bit value.

leolara:
Ok, so perhaps BIT_NUMBER should be SPLIT_64BITS

leolara:
A part from that you can merge. Congratulations, this is going to be read by many people.

PR #202, opened 2024-02-11
implemented low to high algorithm for modular exponentiation

rutefig:


src/parser/mod.rs
```diff
@@ -119,4 +119,66 @@ mod test {
...
+                }
+            }
+
+            state middle {
+                if e == 0 {
+                    -> final {
+                        acc' <== acc;
+                    }
+                } else {
+                    if e % 2 == 0 {
```
leolara, line 166: @rutefig so we cannot do `%` or `/` in constaints. But on the right side of `<--` is ok. Like in circom:

https://docs.circom.io/circom-language/constraint-generation/

```
a <-- b/c;
a*c === b;
```

In contraints you can only allow `*`, `+` and  `-`.

So you probably need to create other signals and do something like the code above.

It is important to differentiate the witness generation from the constraints.

rutefig: so I added an internal signal now for that state, assigned the value with the division and then constrained using a quadratic expression

leolara: Yes, but it is the same thing for the modulus. The if makes the condition part of the constraint.

leolara (changes requested):

rutefig:
Fixed discussed mistakes, let me know if there is anything else to improve @leolara

leolara (changes requested):

src/parser/mod.rs
```diff
@@ -119,4 +119,72 @@ mod test {
...
+            }
+
+            state middle {
+                signal div;
+                signal new_acc;
+
+                div <-- e / 2;
+                div * 2 === e;
+
+                if e % 2 {
```
leolara, line 167: The modulus is like the division, we need another signal and emulate it, you cannot have this operation if inside there is a constraint

rutefig: Can you check if it seems right with the new changes pls?

src/parser/mod.rs
```diff
@@ -119,4 +119,75 @@ mod test {
...
+                    acc', a', e' <== acc, a, e;
+                }
+            }
+
+            state middle {
+                signal div;
+                signal new_acc;
+
+                div <-- e / 2;
+                div * 2 === e;
```
leolara, line 165: @rute aren't these both constraitns contradictory?

div * 2 === e;
div * 2 + rem === e;

leolara: Following the circom operators https://docs.circom.io/circom-language/basic-operators/#arithmetic-operators

We should use `\` for this division

rutefig:
I think I was overcomplicating, can you check now @leolara pls?

leolara:
Sorry, you cannot have a var in an if, if that if contain assertions. Because of how ifs are translatated to one PI in the ABEPI transtalator. The previous way was the way to go.

leolara (approved):

PR #203, opened 2024-02-14
Prototype semantic analyser

leolara:


src/compiler/semantic/analyser.rs
```diff
@@ -0,0 +1,328 @@
...
+
+                RULES.apply_new_symbol_statement(self, param, &id.id, &sym);
+
+                self.symbols
+                    .add_symbol(&self.cur_namespace, id.id.name(), sym);
+            }),
+            _ => unreachable!("parser should only produce signals and vars"),
+        });
+
+        result.iter().for_each(|param| match param {
```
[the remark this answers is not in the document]
leolara: It could be, it would make the code more readable.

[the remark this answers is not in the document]
rutefig: and even the name of the function, maybe if we do not split for now we should inform we are analysing also the result here

src/compiler/semantic/analyser.rs
```diff
@@ -0,0 +1,331 @@
...
+            }),
+            // State decl symbols are added in
+            // add_state_decls
+            Statement::StateDecl(_, _, _) => {}
+            _ => {}
+        }
+
+        RULES.apply_statement(self, &stmt);
+
+        // Recursive
```
rutefig, line 221: should we, in the future, worry about this being recursive?

src/compiler/semantic/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+    Machine,
+    State,
+}
+
+/// Information about a symbol
+#[derive(Clone, Debug)]
+pub struct SymTableEntry {
+    definition_ref: DebugSymRef,
+    category: SymbolCategory,
+    ty: Option<String>,
```
rutefig, line 48: What is the `ty`?

leolara: Type, it is a reserved word so usually in rust I saw ty used but I will add comment

src/compiler/semantic/analyser.rs
leolara: Recursive is good, it is just to clarify the different steps that this method makes, perhaps I should split in two new methods

leolara:
@rutefig check again

rutefig (changes requested):
Left some comments just because of the renaming from `namespaces` to `scopes`. There are some leftovers of `namespaces` with `ns`. Other than that everything looks good and right the comments and even the split on the functions we discussed make the code way more clear! Thank you and great job

leolara:
@rutefig check again

rutefig (approved):
All good! Gracias

PR #211, opened 2024-02-27
Implement todos abepi

rutefig:
WIP: implementing if_then_else
Issue: #209

### DIFF
--- a/src/compiler/abepi.rs
+++ b/src/compiler/abepi.rs
@@ -224,7 +224,21 @@ impl<F: From<u64> + Into<u32> + Clone, V: Clone> CompilationUnit<F, V> {
         _lhs: Expression<F, V>,
         _rhs: Expression<F, V>,
     ) -> CompilationResult<F, V> {
-        todo!()
+        assert!(_lhs.is_arith());
+        assert!(_rhs.is_arith());
+
+        let lhs = self.compile_expression_airth(_lhs);
+        let rhs = self.compile_expression_airth(_rhs);
+
+        // In One Zero: 0 is false
+        // lhs != rhs => lhs - rhs == 0F
+        let expr = lhs - rhs;
+
+        CompilationResult {
+            dsym: _dsym,
+            anti_booly: expr.cast_anti_booly(),
+            one_zero: expr.clone(),
+        }
     }
 
     fn compile_expression_and(
--- a/src/compiler/abepi.rs
+++ b/src/compiler/abepi.rs
@@ -396,8 +396,7 @@ impl<F: From<u64> + Into<u32> + Clone, V: Clone> CompilationUnit<F, V> {
             .map(|(if_then, if_else)| {
                 // The AND of the two results
                 // For the OneZero if we have a 0F in any of the two results, the result will be 0F
-                // For the AntiBooly if we have a >0F in any of the two results, the result will be
-                // >0F with the sum of the two >0F
+                // For the AntiBooly we cast the one_zero result
                 let one_zero = if_then.one_zero * if_else.one_zero;
                 let anti_booly = one_zero.cast_anti_booly();
 
--- a/src/compiler/abepi.rs
+++ b/src/compiler/abepi.rs
@@ -399,7 +399,7 @@ impl<F: From<u64> + Into<u32> + Clone, V: Clone> CompilationUnit<F, V> {
                 // For the AntiBooly if we have a >0F in any of the two results, the result will be
                 // >0F with the sum of the two >0F
                 let one_zero = if_then.one_zero * if_else.one_zero;
-                let anti_booly = if_then.anti_booly + if_else.anti_booly;
+                let anti_booly = one_zero.cast_anti_booly();
 
                 CompilationResult {
                     dsym: dsym.clone(),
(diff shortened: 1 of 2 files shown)


src/compiler/abepi.rs
```diff
@@ -218,13 +218,29 @@ impl<F: From<u64> + Into<u32> + Clone, V: Clone> CompilationUnit<F, V> {
...
     ) -> CompilationResult<F, V> {
-        todo!()
+        assert!(_lhs.is_arith());
+        assert!(_rhs.is_arith());
+
+        let lhs = self.compile_expression_airth(_lhs);
+        let rhs = self.compile_expression_airth(_rhs);
+
+        // In One Zero: 0 is false
+        // lhs != rhs => lhs - rhs == 0 is false
```
leolara, line 236: If you substract two different numbers the result is 0?

src/compiler/abepi.rs
```diff
@@ -262,11 +278,37 @@ impl<F: From<u64> + Into<u32> + Clone, V: Clone> CompilationUnit<F, V> {
...
+        let one_zero = sub
+            .iter()
+            .skip(1)
+            .fold(sub[0].one_zero.clone().one_minus(), |acc, se| {
+                acc * se.one_zero.clone().one_minus()
+            });
+
+        CompilationResult {
+            dsym,
+            anti_booly: one_zero.clone(),
```
leolara, line 309: @rutefig perhaps there is an simpler way of doing the anti_booly version based on anti_booly values.

Just multiplication, if one is 0T, the result will be 0T. If all are 0>T, the result will be 0>T. Because we re in a prime field multiplications of 0> will always be 0>.

leolara:
The code is correct, the comments are not very clear. Like: `lhs != rhs => lhs - rhs == 0` is not correct

leolara:
We don't need to implement if then else in this PR, it can come in another PR.

src/compiler/abepi.rs
rutefig: The comment was not so clear, fixed it

src/compiler/abepi.rs
rutefig: We will leave it for later - compare approaches to check which one is more performant

leolara:
@rutefig so the only thing this PR needs is a better explanation in the comment about the translation, then we can merge

src/compiler/abepi.rs
```diff
@@ -224,7 +224,23 @@ impl<F: From<u64> + Into<u32> + Clone, V: Clone> CompilationUnit<F, V> {
...
         _rhs: Expression<F, V>,
     ) -> CompilationResult<F, V> {
-        todo!()
+        assert!(_lhs.is_arith());
+        assert!(_rhs.is_arith());
+
+        let lhs = self.compile_expression_airth(_lhs);
+        let rhs = self.compile_expression_airth(_rhs);
+
+        // In One Zero: 0 is false, 1 is true
```
leolara, line 233:
```
// lhs != rhs is equivalent to not( lhs == rhs)
// In OneZero not(A) is 1 - A. And lhs == rhs is (lhs-rhs).cast_one_zero().
// Hence OneZero expresion is 1 - (lhs-rhs).cast_one_zero()
// In AntiBooly, not(A) = A.one_zero; And before we saw that lhs == rhs in OneZero is 1 - (lhs-rhs).cast_one_zero()
```

src/compiler/abepi.rs
```diff
@@ -262,11 +278,37 @@ impl<F: From<u64> + Into<u32> + Clone, V: Clone> CompilationUnit<F, V> {
...
+        flatten_bin_op(BinaryOperator::Or, lhs, rhs, &mut sub);
+        assert!(sub.iter().all(|se| se.is_logic()));
+
+        let sub = sub
+            .iter()
+            .map(|se| self.compile_expression_logic(se.clone()))
+            .collect::<Vec<_>>();
+
+        // By De Morgan's law, a or b = not (not a and not b)
+        // In OneZero 0F 1T
```
leolara, line 296:
```
// In OneZero not(A) is 1-A. And A and B is A * B. Hence A or B is 1 - ((1-A)*(1-B)).
// For AntiBooly, we can cast to anti-booly the OneZero expression.
```

leolara:
@rutefig please, check my version of the comments, by going step by step you can catch errors in the reasoning.

src/compiler/abepi.rs
```diff
@@ -323,12 +365,55 @@ impl<F: From<u64> + Into<u32> + Clone, V: Clone> CompilationUnit<F, V> {
...
+
+        // if A then assert B else assert C
+        // this is equivalent to (if A then assert B) and (if not A then assert C)
+        // this will be equivalent to (not (A and not B)) and (not (not A and not C))
+
+        // Using cond only as OneZero 0F, 1T
+        // For the OneZero result and we can write this as
+        // (1 - (cond * (1-when_true.one_zero))) * (1 - ((1-cond) * (1-when_false.one_zero)))
+        // For the AntiBooly result we can do
+        // cond * when_true.anti_booly * (1-cond) * when_false.anti_booly
```
leolara, line 383: @rutefig the explanation of the anti-booly has a * where in the code there is a +.

Also, if you are reusing the other code, you don't need to convert to formula the reuse. The only thing you are doing outside the reuse, is the AND of `(if A then assert B) and (if not A then assert C)`. If you are implementing it like this, you only need to explain how you are doing the AND.

leolara (approved):

PR #217, opened 2024-02-28
implemented rule to constraint assignments only to signals

rutefig:
Implemented the rule to only allow assignments (`<--` and `<==`) to signals and did a simple test to check if we throw the error when we are doing that to a wg var


leolara:
Perfect!! The only thing perhaps the message, because = will allow assignment to wgvars. So perhaps we can say that you cannot use <-- but can use = or something like that

leolara (approved):

Issue #216, opened 2024-02-28
Add missing arithmetic operators to parser


leolara:
We should have all that are in section "Arithmetic operators" and "Bitwise operators" here https://docs.circom.io/circom-language/basic-operators/#arithmetic-operators

The operator priority should be like in circom that is the same as in rust.

rutefig:
In this branch: https://github.com/privacy-scaling-explorations/chiquito/tree/rutefig/205-add-missing-binary-operators-to-lalrpop-and-ast
It is linked to the issue #205

leolara:
@rutefig why do you link to a branch instead of sending a PR?

rutefig:
> @rutefig why do you link to a branch instead of sending a PR?

It was just to keep track that this issue is being done on the same branch as the issue #205 because it is related, when the PR is ready I will link it here too

Issue #205, opened 2024-02-20
Add missing binary operators to LALRPOP and AST


leolara:
They are not compiled by the ABEIP

PR #224, opened 2024-03-08
add missing binary operators to lalrpop and ast

rutefig:
Added missing logic, arithmetic and bitwise operators to lalrop and ast.

Covers #205 and #216 issues

### DIFF
--- a/src/parser/chiquito.lalrpop
+++ b/src/parser/chiquito.lalrpop
@@ -136,6 +136,11 @@ ParseEq: Expression<BigInt, Identifier> = {
     <lhs: ParseArith> "==" <rhs: ParseArith> => build_bin_op("==", lhs, rhs),
     <lhs: ParseArith> "!=" <rhs: ParseArith> => build_bin_op("!=", lhs, rhs),
 
+    <lhs: ParseArith> "<" <rhs: ParseArith> => build_bin_op("<", lhs, rhs),
+    <lhs: ParseArith> "<=" <rhs: ParseArith> => build_bin_op("<=", lhs, rhs),
+    <lhs: ParseArith> ">" <rhs: ParseArith> => build_bin_op(">", lhs, rhs),
+    <lhs: ParseArith> ">=" <rhs: ParseArith> => build_bin_op(">=", lhs, rhs),
+
     ParseArith
 }
 
--- a/src/parser/chiquito.lalrpop
+++ b/src/parser/chiquito.lalrpop
@@ -152,34 +152,36 @@ ParseBitAnd: Expression<BigInt, Identifier> =  ParseLogichBinOp<"&", ParseArith>
 
 // Arithmetical expressions
 ParseArith: Expression<BigInt, Identifier> = {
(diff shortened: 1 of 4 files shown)


src/parser/chiquito.lalrpop
```diff
@@ -136,21 +136,52 @@ ParseEq: Expression<BigInt, Identifier> = {
...
+ParseBitAnd: Expression<BigInt, Identifier> =  ParseLogichBinOp<"&", ParseArith>;
+
 // Arithmetical expressions
 ParseArith: Expression<BigInt, Identifier> = {
-    ParseSum
+    ParseLeftShift
 }
 
-ParseSum: Expression<BigInt, Identifier> = ParseArithBinOp<"+", ParseMul>;
+ParseLeftShift: Expression<BigInt, Identifier> = ParseArithBinOp<"<<", ParseRightShift>;
```
leolara, line 158: @rutefig << and >> have the same presence so you need to group them like in ParseEq

src/parser/chiquito.lalrpop
```diff
@@ -136,21 +136,52 @@ ParseEq: Expression<BigInt, Identifier> = {
...
+
+ParseRightShift: Expression<BigInt, Identifier> = ParseArithBinOp<">>", ParseSub>;
```
leolara, line 160: @rutefig group + and -

src/parser/chiquito.lalrpop
```diff
@@ -136,21 +136,52 @@ ParseEq: Expression<BigInt, Identifier> = {
...
+
+ParseSub: Expression<BigInt, Identifier> = ParseArithBinOp<"-", ParseSum>;
+
+ParseSum: Expression<BigInt, Identifier> = ParseArithBinOp<"+", ParseDivRem>;
 
-ParseMul: Expression<BigInt, Identifier> = ParseArithBinOp<"*", ParsePrefixMinus>;
+ParseDivRem: Expression<BigInt, Identifier> = ParseArithBinOp<"%", ParseDiv>;
```
leolara, line 166: @rutefig group * / %, \\

leolara:
The circom documentation references the rust documentation about precendence:
https://docs.circom.io/circom-language/basic-operators/
https://doc.rust-lang.org/1.22.1/reference/expressions/operator-expr.html#operator-precedence

This does not explain the precedence of **, ~ so we need to investigate which is actually in circom

rutefig:
> The circom documentation references the rust documentation about precendence: https://docs.circom.io/circom-language/basic-operators/ https://doc.rust-lang.org/1.22.1/reference/expressions/operator-expr.html#operator-precedence
> 
> This does not explain the precedence of **, ~ so we need to investigate which is actually in circom

Yeah I checked the lalrop file, `~` is a unary operator so it has the same precedence as the rest of unary operators, and pow `**` comes before the mul and div group, right after the unary operators.

Regarding the groups I agree it should all be grouped :) gonna do it right away

leolara (approved):

Issue #227, opened 2024-03-18
Symbol Table: get_type


leolara:
Add method get_type to SymbolTableEntry, if type is None, return "field"

PR #232, opened 2024-03-22
implemented get_type function

rutefig:
deref type and if it's None return "field"


leolara:
Closes #227

leolara (approved):

leolara:
@rutefig Beautiful ❤️ 

Why do you need unwrap?

leolara:
like, you could return `_ty` ? I think

leolara:
I mean ty.as_deref ?

rutefig:
> like, you could return `_ty` ? I think

yes you're right thank you 🙏🏼

PR #234, opened 2024-03-23
Chiquito 2024 to sbpir compiler

leolara:


rutefig (approved):
Just asking if we should leave the prints in the compiler, and if we can have some more comments on the code. But the code itself looks pretty clean and clear too, so I think is not really bad if we don't have all of it commented 👌🏼 

Great work!

leolara:
@rutefig I added comments

PR #236, opened 2024-03-26
implemented rule to check if true and false are being used inside of …

rutefig:
…logical expressions

### DIFF
--- a/src/compiler/semantic/rules.rs
+++ b/src/compiler/semantic/rules.rs
@@ -221,14 +221,12 @@ fn machine_decl_tl(
                             ),
                             dsym,
                         ),
-                        
                     }
                 })
             } else {
                 unreachable!("parser only generates blocks in this context");
             }
         }
-        
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/compiler/semantic/rules.rs
```diff
@@ -251,10 +249,48 @@ fn redeclare_rule(
     }
 }
 
+// Cannot use true or false if it's not inside of a logic expression.
+fn true_false_rule(analyser: &mut Analyser, expr: &Expression<BigInt, Identifier>) {
+    use Expression::*;
```
leolara, line 254: @rutefig I think this does not go recursively and hence only checks at the second level of the Expression AST

rutefig: True, fixing it

src/compiler/semantic/rules.rs
```diff
@@ -251,10 +249,49 @@ fn redeclare_rule(
...
+
+    match expr {
+        BinOp { lhs, rhs, .. } => {
+            check_true_false(analyser, expr, lhs);
+            check_true_false(analyser, expr, rhs);
```
leolara, line 259: @rutefig I think that the recursion into true_false_rule is done inside check_true_false is confusing. It is easier to read if true_false_rule calls itself after calling check_true_false.

leolara:
```
BinOp { lhs, rhs, .. } => {
            check_true_false(analyser, expr, lhs);
            check_true_false(analyser, expr, rhs);
            true_false_rule(analyser, lhs);
            true_false_rule(analyser, rhs);
```

leolara (approved):

PR #239, opened 2024-04-30
Rutefig/231 research common subexpression elimination

rutefig:
Added the generic `M` as discussed to store metadata, and where it will also store the hash of each expression node to apply the CSE optimization.

Anything I should improve just lmk as I'm not so confident with generics in rust :) thank you


src/compiler/abepi.rs
```diff
@@ -13,29 +13,31 @@ use crate::{
 
 #[derive(Debug)]
 /// Result of compiling an arbitrary boolean expression into a PI
-pub struct CompilationResult<F, V> {
+pub struct CompilationResult<F, V, M> {
     #[allow(dead_code)]
     pub dsym: DebugSymRef,
     // 0 is true, !=0 is false
-    pub anti_booly: Expr<F, V>,
+    pub anti_booly: Expr<F, V, M>,
```
leolara, line 20: I think we talked about this, here it would be `pub anti_booly: Expr<F, V, ()>` so there is no need to add M to CompilationResult. And it is true for the rest, Everywhere that Expr is used M is set to `()`.

rutefig: I'm sorry, I was just thinking in moving forward on applying already `M` where I was thinking it would be necessary.
Fixing it

src/compiler/compiler.rs
```diff
@@ -25,7 +25,7 @@ use super::{
 
 /// This compiler compiles from chiquito source code to the SBPIR.
 #[derive(Default)]
-pub(super) struct Compiler<F> {
+pub(super) struct Compiler<F, M> {
```
leolara, line 28: @rutefig Again I don't think we need to add M, or the phantom data

src/poly/mod.rs
```diff
@@ -13,28 +13,30 @@ pub mod mielim;
...
+    Mul(Vec<Expr<F, V, M>>),
+    Neg(Box<Expr<F, V, M>>),
+    Pow(Box<Expr<F, V, M>>, u32),
     Query(V),
     Halo2Expr(Expression<F>),
 
-    MI(Box<Expr<F, V>>), //  Multiplicative inverse, but MI(0) = 0
+    MI(Box<Expr<F, V, M>>), //  Multiplicative inverse, but MI(0) = 0
+
+    Metadata(M),
```
leolara, line 36: @rutefig what is this for?

M should be added to all the fields, like `Sum(Vec<Expr<F, V, M>>, M)`

src/compiler/mod.rs
```diff
@@ -56,7 +56,7 @@ pub fn compile<F: Field + Hash>(
     source: &str,
     config: Config,
 ) -> (Result<SBPIR<F, ()>, ()>, Vec<Message>) {
-    let mut compiler = Compiler::new(config);
+    let mut compiler: Compiler<F> = Compiler::new(config);
```
leolara, line 59: @rutefig the type can be infer by the compiler

src/poly/mod.rs
leolara: So each of the nodes can have metainfo, so a additional field of type M is added to all of them

rutefig: Yes I was divided into doing that or doing a separate variant because it simplifies the integration, but then we needed to ensure each node had this field. I was just moving forward with one type of solution knowing it could be not the one you had imagined. So I'm going to rewind and move with the one you said

leolara: Yes, is that Metadata(M) I don't think has an use. I know it is a lot of repetitibve work, but we are going to use it in many places.

rutefig: Yeah it's okay, did the change :)

src/poly/mod.rs
```diff
@@ -13,43 +13,43 @@ pub mod mielim;
...
 }
 
 pub trait ToField<F> {
     fn field(&self) -> F;
 }
 
 #[derive(Clone)]
-pub enum Expr<F, V> {
+pub enum Expr<F, V, M> {
     Const(F),
```
leolara, line 26: @rutefig const and query will also need metainfo

leolara (approved):
Please, run the example before merging to make sure the stay working

PR #243, opened 2024-05-20
242 make debugsymref usable

alxkzmn:


src/compiler/circuit_error.zk
```diff
@@ -0,0 +1,32 @@
...
+            -> final {
+                i', b', n' <== i + 1, c, n;
+            }
+        } else {
+            -> middle {
+                i', a', b', n' <== i + 1, b, c, n;
+            }
+        }
+    }
+}
```
leolara, line 32: empty end of line

src/compiler/circuit_error.zk
```diff
@@ -0,0 +1,32 @@
+machine fibo(signal n) (signal b: field) {
```
leolara, line 1: perhaps ".chiquto" and put it in another folder like ./test or name the file `test_`

src/parser/ast/debug_sym_factory.rs
```diff
@@ -0,0 +1,78 @@
+use codespan_reporting::files::{Files, SimpleFile};
```
leolara, line 1: super good catch, I didn't know about this library and I thought we had to build this

leolara:
You have done more than the, task, I was going to put calculating the line and col in anther task!!

Do you think it is a good optimization to put the file name in Rc<String> or Rc<str> ?

src/compiler/circuit_error.zk
alxkzmn: Sorry, I don't get it

alxkzmn:
@leolara thanks! :) I found that crate because I was curious how others are logging errors, so that we don't reinvent the wheel. What is "Rc"?

leolara:
@alxkzmn I mean this https://doc.rust-lang.org/std/rc/struct.Rc.html so we don't copy the filename string.

alxkzmn:
Sure, that's a better way!

leolara:
@alxkzmn aparently Rc<str> is more efficient https://www.reddit.com/r/rust/comments/qlbl3v/rcstr_vs_rcstring_for_interpreter/

leolara (approved):

leolara:
@alxkzmn can you merge?

PR #246, opened 2024-05-21
Rutefig/cse hash expressions

rutefig:
Hash function for `Expr` nodes using Schwartz-Zippel Lemma: takes an array of variable assignments and hashes the expression based on the evaluation for every assignments

### DIFF
--- a/src/poly/cse.rs
+++ b/src/poly/cse.rs
@@ -68,7 +68,10 @@ mod test {
 
     use halo2_proofs::halo2curves::bn256::Fr;
 
-    use crate::sbpir::{query::Queriable, ForwardSignal, InternalSignal};
+    use crate::{
+        poly::{Expr::*, ToExpr},
+        sbpir::{query::Queriable, ForwardSignal, InternalSignal},
+    };
 
     #[test]
     fn test_generate_random_assignment() {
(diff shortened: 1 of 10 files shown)


src/poly/mod.rs
```diff
@@ -1,7 +1,7 @@
 use std::{
     collections::HashMap,
     fmt::Debug,
-    hash::Hash,
+    hash::{DefaultHasher, Hash, Hasher},
```
alxkzmn, line 4: Seems like Hasher is not used.

src/poly/mod.rs
```diff
@@ -47,6 +47,60 @@ impl<F, V, M> Expr<F, V, M> {
...
+
+#[derive(Debug, Clone)]
+pub struct HashResult {
+    pub hash: u64,
+    pub degree: usize,
+}
+
+impl<F: Field + Hash, V: Debug + Clone + Eq + Hash> Expr<F, V, ()> {
+    /// Uses Schwartz-Zippel Lemma to hash
+    pub fn hash(&self, assignments: &[VarAssignments<F, V>]) -> Expr<F, V, HashResult> {
```
alxkzmn, line 73: What do you think of passing a shared instance of hasher into all the expressions upon their creation? In that case we would be sure that all hashes are evaluated at the same point without the need to pass assignments as an argument.

src/poly/mod.rs
rutefig: Hasher is a trait that implements `finish` that is being used from `DefaultHasher`

src/poly/mod.rs
rutefig: I think it makes sense, thanks for the observation 🙏🏼

src/poly/mod.rs
alxkzmn: Got it!

src/poly/mod.rs
alxkzmn: I'll close the discussion for now as we're not sure whether we would call the hash function in more than one place across the project. We can keep that in mind for the future.

alxkzmn (approved):
LGTM!

src/poly/mod.rs
rutefig: As discussed for now it will be implemented with the array of assignments, later on we can check if we need to use a var manager and pass the reference to the hasher instead

rutefig:
> @alxkzmn it might seem I wasn't reviewing this, but I wrote Rute in chat and ask her to add a function that gets a list of queriables and returns a random point. Now that we are having more people I should have wrote here

I also mentioned in the daily that you have reviewed too :) and I'm moving forward to implement the random generation in this same PR

src/poly/cse.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+fn generate_random_assignment<F: Field + Hash, V: Debug + Clone + Eq + Hash>(
+    keys: HashSet<V>,
+) -> VarAssignments<F, V> {
+    let mut rng = ChaCha20Rng::seed_from_u64(0);
+
+    let mut assignments = VarAssignments::new();
+
+    for key in keys.iter() {
```
leolara, line 42: For forward signals we should add one random for the rotation 0 and another for rotation 1

src/poly/cse.rs
```diff
@@ -0,0 +1,83 @@
...
+
+use crate::field::Field;
+
+use std::{collections::HashSet, fmt::Debug, hash::Hash};
+
+use super::{Expr, HashResult, VarAssignments};
+
+/// Common Subexpression Elimination - takes a collection of expr
+pub fn cse<F: Field + Hash, V: Debug + Clone + Eq + Hash>(
+    exprs: Vec<Expr<F, V, ()>>,
```
leolara, line 11: I think we can pass here the queriables, because the caller will have the list of all queriables and it will be more efficient than calling get_queries in all the expr

rutefig: Can we use https://github.com/privacy-scaling-explorations/chiquito/blob/93282e00b925fd3d61ea32c0f48fd8385f385b51/src/sbpir/mod.rs#L28C9-L28C20 directly instead of the queriables?

src/poly/cse.rs
rutefig: Why shouldn't they use the same assignment? Because in the end they are like the same value being used in two different PI's right? Or maybe I'm just not making any sense 😅

src/poly/cse.rs
leolara: The annotations? We need to know, if they are internal or forward. You could either pass queriables, or an array of internals and another of forward

src/poly/cse.rs
leolara: The assignments are random, there show me two random assignations to each forward, one for this step and another for the next. Because there can be queries to the next step (Queriable::Forward(_, true))

src/poly/cse.rs
```diff
@@ -0,0 +1,77 @@
...
+
+fn generate_random_assignment<F: Field + Hash>(
+    queriables: &Vec<Queriable<F>>,
+) -> VarAssignments<F, String> {
+    let mut rng = ChaCha20Rng::seed_from_u64(0);
+
+    let mut assignments = VarAssignments::new();
+
+    for queriable in queriables {
+        let value = F::random(&mut rng);
```
leolara, line 35: If the queriable is Forward, it should produce two values, one with false and another with true in `.1`

leolara (approved):

PR #252, opened 2024-05-29
WG interpreter

leolara:


src/compiler/compiler.rs
```diff
@@ -151,25 +173,31 @@ impl<F: Field + Hash> Compiler<F> {
...
-                                ctx.constr(constraint);
+                    ctx.step_type_def(
+                        self.mapping.get_step_type_handler(machine_id, state_id),
+                        |ctx| {
+                            self.add_internals(ctx, symbols, machine_id, state_id);
+
+                            ctx.setup(|ctx| {
+                                let pis =
+                                    self.translate_queries(symbols, setup, machine_id, state_id);
+                                pis.iter().for_each(|pi| {
```
alxkzmn, line 184: What does pi stand for?

src/compiler/compiler.rs
```diff
@@ -151,25 +173,31 @@ impl<F: Field + Hash> Compiler<F> {
...
+                                    let constraint = Constraint {
+                                        annotation: format!("{:?}", pi),
+                                        expr: pi.clone(),
+                                        typing: Typing::AntiBooly,
```
alxkzmn, line 188: AntiNooly is a funny name :) Does it mean a negated boolean?

src/interpreter/mod.rs
```diff
@@ -0,0 +1,492 @@
...
+impl<'a, F: Field + Hash> Interpreter<'a, F> {
+    pub fn new(symbols: &'a SymTable, mapping: &'a SymbolSignalMapping) -> Self {
+        Self {
+            mapping,
+            cur_frame: StackFrame::new(symbols),
+            witness: Vec::default(),
+        }
+    }
+
+    pub fn run(
```
alxkzmn, line 38: Let's add docs to public functions. Is the input a mapping of input signal names to values? (that's my guess)

src/compiler/compiler.rs
leolara: polynomial identity, like constraint, it is true that it is a bit obscure

src/compiler/compiler.rs
leolara: Booly is boolean like in C, 0 is false anything else is true. Anti-booly is the oposite, 0 is true, anything else is false. Then we have OneZero (1 true, 0 false, it is not possible something else).

To convert arbitrary boolean expressions into PI we need to use both represenations and do casting between them. Casting OneZero to Anti-Booly is 1-x; The other way around is the is_zero gadget

src/interpreter/mod.rs
```diff
@@ -0,0 +1,492 @@
...
+            None
+        }
+    }
+}
+
+pub fn run<F: Field + Hash>(
+    ast: &[TLDecl<BigInt, Identifier>],
+    symbols: &SymTable,
+    mapping: &SymbolSignalMapping,
+    input: HashMap<String, F>,
```
alxkzmn, line 246: Same here, is input a mapping of signals to values?

src/wit_gen.rs
```diff
@@ -278,6 +282,111 @@ impl<F: Field + Hash> FixedGenContext<F> {
...
+    pub fn generate<F: Field + Hash>(
+        &self,
+        input: HashMap<String, F>,
+    ) -> Result<TraceWitness<F>, Message> {
+        let step_instances = interpreter::run(&self.ast, &self.symbols, &self.mapping, input)?;
+
+        Ok(TraceWitness { step_instances })
+    }
+
+    pub fn evil_assign<F: Field + Hash, S: Into<String>>(
```
alxkzmn, line 375: Does it mean something like assign invalid?

src/interpreter/mod.rs
```diff
@@ -0,0 +1,492 @@
...
+            SingleRowCellManager {},
+            SimpleStepSelectorBuilder {},
+        ));
+
+        let compiled = chiquito2Halo2(plonkish.0);
+
+        let circuit = ChiquitoHalo2Circuit::new(
+            compiled,
+            plonkish.1.map(|g| g.generate_with_witness(witness)),
+        );
```
alxkzmn, line 403: Why do we need an intermediate `let compiled = chiquito2Halo2(plonkish.0);` instead of going from `plonkish.0` to `circuit` directly?

src/compiler/compiler.rs
alxkzmn: Let's rename it then :)

src/compiler/compiler.rs
alxkzmn: It would be cool to have it as a doc above the type enum.

alxkzmn (changes requested):
Let's add docs to the new public functions.

src/compiler/compiler.rs
leolara: to what do you think we should rename it?

src/compiler/compiler.rs
leolara: there is this, but I think there are some mistakes in the code

https://github.com/leolara/notes/blob/main/drafts/eugenio.md

src/wit_gen.rs
leolara: Yes, this is a method where you can change the value of any singal in the witness. It is useful for finding soundness bugs and to test that the circuit generates an invalid proof for an invalid witness.

In zkEVM @ChihChengLiang called this "evil witness or evil prover"

src/interpreter/mod.rs
leolara: Yes, this is design decision from the fact that you need the witness to create a halo2 circuit, when we compile to halo2 midleware perhaps we can make this interface more natural

src/compiler/compiler.rs
alxkzmn: poly_identity

src/interpreter/mod.rs
leolara: I don't understand this question, it is a mapping of identifiers to values. Do you mean that it should be commented?

leolara:
Please,. check again @alxkzmn

src/interpreter/mod.rs
alxkzmn: Yes

alxkzmn (approved):

alxkzmn:
Thanks for all the comments! All looks good to me.

leolara:
@alxkzmn after the TraceGenerator refactoring I am going to refactor this to use that

PR #254, opened 2024-05-30
Changes for the language server

alxkzmn:
The following changes were necessary to integrate with the language server:

- Usages tracking for `SymTableEntry` (and some WIP additional functionality in the Analyzer to record the usages during the analysis).
- "Reverse" lookup by offset in `SymTable`
- File reference in DebugSymRef changes from Rc to Arc because the language server is asynchronous.


alxkzmn:
I see that some tests are failing due to the new `usages` field in the SymTableEntry, but I'd rather fix them only if the changes make it through the review :)

src/compiler/semantic/analyser.rs
```diff
@@ -243,6 +261,17 @@ impl Analyser {
             // State decl symbols are added in
             // add_state_decls
             Statement::StateDecl(_, _, _) => {}
+            Statement::Transition(dsym_ref, id, _) => {
+                // Find the corresponding symbol and add usage
+                if let Some(entry) = self.symbols.find_symbol(&self.cur_scope, id.name()) {
```
leolara, line 266: @alxkzmn 
 
Wouldn't find_symbol return `None` if the symbol has not been declared before?

I think the best solution for the problem of usage before declaration is to add the usages as a second pass, after the symbol table has been created. That also has the advantege that if we don't need to use the usages in the compiler we don't have to run it. Also you could create it as a different function or struct from the semantic analyser. Perhaps `ProcessUsages` or something like that.

src/compiler/semantic/mod.rs
```diff
@@ -286,6 +290,68 @@ impl SymTable {
...
+            // Return the first symbol in the map because BTreeMap is sorted by the key (which is
+            // the proximity in our case)
+            return symbols_by_proximity
+                .iter()
+                .next()
+                .map(|(_, entry)| entry.clone());
+        }
+    }
+
+    fn look_in_usages(
```
leolara, line 338: @alxkzmn it seems this method doesn't use `&self` but `&SymTableEntry` perhaps should be a method in `SymTableEntry`

src/parser/ast/debug_sym_factory.rs
```diff
@@ -1,13 +1,13 @@
...
-use codespan_reporting::files::{Files, SimpleFile};
+use codespan_reporting::files::SimpleFile;
 
 use super::DebugSymRef;
 
 /// Factory for creating debug symbol references.
 pub struct DebugSymRefFactory {
     /// Source file reference.
-    file: Rc<SimpleFile<String, String>>,
+    file: Arc<SimpleFile<String, String>>,
```
leolara, line 10: @alxkzmn you introduced `Arc` in case we are doing multi-threading compilation?

src/parser/ast/mod.rs
```diff
@@ -11,32 +11,81 @@ pub mod tl;
...
-            line_start,
-            col_start,
-            line_end,
-            col_end,
-            file,
+    pub fn new(start: usize, end: usize, file: Arc<SimpleFile<String, String>>) -> DebugSymRef {
+        DebugSymRef { start, end, file }
+    }
+
+    fn get_column_number(&self, line_index: usize, start: usize) -> usize {
```
leolara, line 27: @alxkzmn why do you think it is better to don't pre-compute the values? For performance?

src/parser/ast/debug_sym_factory.rs
alxkzmn: No, this is because the language server is inherently asynchronous, so if it shares the code with the comoiler it won't compile with Rc.

src/parser/ast/mod.rs
alxkzmn: Because the things like rows and columns are file-related so they are needed only in user-facing scenarios like language server.

src/parser/ast/debug_sym_factory.rs
leolara: Ok :+1:

src/parser/ast/mod.rs
leolara: Ok 👍  we will use them in the error messages, but I guess we only need to calculate them for those, so it is good

src/compiler/semantic/analyser.rs
alxkzmn: Totally agree, right now it's not working for states that are declared later down the code. I'll refactor.

alxkzmn: - I see a problem with the "second pass" approach that stems out of our AST being non-uniform. What I mean by non-uniform is that the only way to traverse the tree is to know exactly what each level looks like. For example, the top-level declaration is a machine, and machine has a `block` of statements, and statements are all different types as well. For the "second pass" approach it means repeating a lot of code that descends the AST in the same way as the analyzer, through a series of custom `match` and analysis statements specific for each level of the tree. In such a case the language server code will have to repeat the custom tree traversal from the analyzer code.
- Another approach could be to make each level of AST implement a common trait like `Node` that has `children: Vec<Node>` during parsing. This way tree traversal becomes uniform regardless of the level - you can recursively go through children and look for any necessary information. This would allow the language server to be independent and not have to copy all the custom `match`-es (and maybe would enhance the analyzer as well, for example, by adding a `parent: Node` to each `Node` and then being able to look up and down the scope(s)).
- The hybrid approach in the middle of the spectrum is to do everything in the same place (record usages and other necessary info while analyzing) which is in this PR. Here I understand your concern that not all of the code would be necessary for the end goal of compiling the circuit.

leolara: It would have been done with `Vec<Node>` in another language like Java. However, in rust `Vec<Node>` is not legal if `Node` is a trait. You could do `Vec<Box<dyn Node>>` but in my experience this has been very problematic what I have done it in rust. That is the reason it is implemented like this currently. Change it to  `Vec<Box<dyn Node>>` would be a lot of work.

For me what is problematic from your approach is that you are adding symbols in different places that were the original code was done and make the code more difficult to mantain because before there was only a place where a symbol was added.

However, now I remember that we found the same problem when building the semantic analyser. The only symbols that can be used before declaring them are the states, in the transitions. That is why we call `add_state_decls` before processing the content of the state.

Hence, the states symbols should be declared before being in any transition statement. The current code should work without adding symbols. I think, so it is even simpler in the end.

Or perhaps I am missing something here

leolara:
so, now the changes to the analyser are missing?

alxkzmn:
@leolara maybe I misunderstood but I thought that the conclusion was to extract the usages in a separate pass and put them in a separate data structure. In this case it could all be done in the language server.

leolara:
@alxkzmn  That is not the conclusion: https://github.com/privacy-scaling-explorations/chiquito/pull/254#discussion_r1625953580

The conclusion is to remove the lines that check if the state symbol is there and create it if it is not. The state symbol will be there.

alxkzmn:
@leolara to make sure I'm on the right track:
- it's OK to process the usages while analyzing, all in single pass;
- the usages should be recorded to a separate data structure.
Is this correct?

leolara:
Yes, single pass, you don't need a separate data structure. What you had before is ok, just needs to assume that the states are already in the symbol table, because they should be. As I explained, we are already doing kind of two passes, one to collect the state symbols and then to analyse each state after. So when you are analysing a transition you already should have the state in the symbol table

alxkzmn:
@leolara the PR is ready for review.

src/compiler/semantic/analyser.rs
```diff
@@ -250,12 +268,18 @@ impl Analyser {
...
         match stmt {
             Statement::Assert(_, expr) => self.analyse_expression(expr),
-            Statement::SignalAssignment(_, _, exprs) => exprs
-                .into_iter()
-                .for_each(|expr| self.analyse_expression(expr)),
-            Statement::SignalAssignmentAssert(_, _, exprs) => exprs
-                .into_iter()
-                .for_each(|expr| self.analyse_expression(expr)),
+            Statement::SignalAssignment(_, ids, exprs) => {
+                self.extract_id_usages(&ids);
```
leolara, line 272: @alxkzmn wouldn't it be better to extract the usages after the semantic analysis? I mean swaping the calls to extract_id_usages and analyse_expression ?

src/compiler/semantic/analyser.rs
```diff
@@ -282,9 +309,43 @@ impl Analyser {
...
+            self.symbols
+                .update_usages(&self.cur_scope, id.name(), id.debug_sym_ref());
+        }
+    }
+
+    fn extract_usages_recursively(&mut self, expr: &Expression<BigInt, Identifier>) {
+        match expr.clone() {
+            Expression::Query(dsym_ref, id) => {
+                self.symbols
+                    .update_usages(&self.cur_scope, id.name(), dsym_ref);
```
leolara, line 327: @alxkzmn why in this case use the dsym of query and on others of the identifier itself?

src/parser/ast/mod.rs
```diff
@@ -11,32 +11,81 @@ pub mod tl;
...
+
+    /// Returns the proximity score of the given offset to the debug symbol.
+    /// The proximity score is the sum of the distance from the start and end of the symbol.
+    /// If the offset is not within the symbol, -1 is returned.
+    pub fn proximity_score(&self, offset: usize) -> i32 {
+        if self.start <= offset && offset <= self.end {
+            let start_diff = offset as i32 - self.start as i32;
+            let end_diff = self.end as i32 - offset as i32;
+
+            start_diff + end_diff
```
leolara, line 86: I think it would be more idiomatic to return Option<u32> Instead of -1

leolara: Could you add tests for this method

src/compiler/semantic/mod.rs
```diff
@@ -78,13 +80,32 @@ impl SymTableEntry {
             None => "field",
         }
     }
+
+    fn look_in_usages(
```
leolara, line 84: This method, for self symbol, find its usages that enclose an offset. Right?

perhaps `find_usages_at` ?

src/parser/ast/mod.rs
leolara: I think if `self.start <= offset <= self.end` then `offset - self.start + self.end - offset == self.start + self.end`. I think the value that you want is `self.end - self.start`. And for that you don't need offset. It seems what you want is the size of the symbol footprint.

src/compiler/semantic/analyser.rs
rutefig: I agree, at least it seems more intuitively to make more sense to analyse first and then extract the usages

alxkzmn: The IDs weren't initially used (_), so shouldn't be any difference (only affects the order in which they are recorded). But I'll rename and move the method after the analysis, just in case.

src/compiler/semantic/analyser.rs
alxkzmn: Let's take a look at how the parser is treating the different cases:

- expressions like `a + b`: `a` and `b` are eventually the queries to the signals so we can use their dsym_ref;
- assignments like `a <-- b + c` - although for a programmer it looks like both left and right side are similar to expressions above, for the parser the left side is an `ID` and the right side is an expression, so the right side will be recursively parsed as `Queries` above, but the left side needs special handling as `ID`s. Same goes for the state transitions like `-> final` - `final` here is an ID.

leolara: Yes, what I was thinking is that to make the code look the same perhaps we could use the dsym of the identifier instead of the query. That in this case are both the same. But it is not important

alxkzmn: @leolara I think I got what you mean, we could refactor the `Query` to not have a separate debug sym ref because it can use the debug sym ref of its `Identifier`.

alxkzmn: The only obstacle for that is that in `Expression<F, V>` V is unbound. What's the motivation behind making V generic? Can it be anything besides `Identifier`?

src/parser/ast/mod.rs
```diff
@@ -11,32 +11,78 @@ pub mod tl;
...
+        self.get_column_number(line_idx, self.end)
+    }
+
+    pub(crate) fn get_filename(&self) -> String {
+        self.file.name().to_string()
+    }
+
+    /// Returns the proximity score of the given offset to the debug symbol.
+    /// The proximity score is the sum of the distance from the start and end of the symbol.
+    /// If the offset is not within the symbol, -1 is returned.
```
leolara, line 80: The comment is no longer valid

leolara:
I put all my ideas here together to avoid more cycles:
 + It seems update_usages is repeating what find_symbol does? Cannot be it rewritten using find_symbol?
 + Here we are reading and "adding" a symbol to update its usages. Wouldn't it be better to add a method add_usage for this, more readable and encapsulated (https://github.com/privacy-scaling-explorations/chiquito/pull/254/files#diff-feef6fedaf732a222ce51a0e815b9afe9d0386c1099a96d97cd3b6dca7ba8f60R291-R299)
  + In `find_symbol_by_offset` I think the if can be simplified by putting an `and` of the first two conditions. (https://github.com/privacy-scaling-explorations/chiquito/pull/254/files#diff-feef6fedaf732a222ce51a0e815b9afe9d0386c1099a96d97cd3b6dca7ba8f60R374-R376)
  + In  `check_usage_at` the code can be simplified and more stateless by returning only the "closest" of the collected symbols. This can be done without a map. Keeping the closest found until now in a variable.
  + In `find_symbol_by_offset` as well the code can be simplified by using a variable to keep the closest found until now, and don't use a map.
   + `extract_usages_recursively` is better rename to `extract_usages_expression` and be called after the analysis like in the statements. This is more coherent with the current naming.
   + `extract_usages_expression` could call `collect_id_usages([id])` but perhaps this is too nit pick.

src/compiler/semantic/mod.rs
```diff
@@ -289,6 +354,37 @@ impl SymTable {
...
+    /// Find a `SymTableEntry` by its byte offset in a file.
+    /// The function can be called externally (e.g., from the language server)
+    ///
+    /// ### Parameters
+    /// - `filename`: The name of the file where the symbol is searched.
+    /// - `offset`: The byte offset in the file where the symbol is searched.
+    /// ### Returns
+    /// The `SymTableEntry` that is closest to the offset.
+    pub fn find_symbol_by_offset(&self, filename: String, offset: usize) -> Option<SymTableEntry> {
+        let mut closest_symbol: Option<SymTableEntry> = None;
```
leolara, line 367: @alxkzmn I think it would be much more efficient and clear to store the closest_symbol_proximity in a variable instead of calling proximity score every time.

alxkzmn: I initially wanted to do that but then refactored into calling the method for readability because it seemed like overoptimizing.

leolara: I think it also make it more readbale, also perhaps you could have an and in the two conditions, although I think that feature with `if let` is experimental.

But I also thought for a moment that proximity_score had a couple of loops inside so I thought it was more important.

leolara (approved):

leolara:
It was aprove already, you can merge

PR #256, opened 2024-06-02
implemented if expression should be bool rule

rutefig:


alxkzmn (approved):

leolara:
The test should pass if the expression is a query on a signal that is of type bool. I am not sure if this works with is_arith is_bool

leolara:
And as oppositte we should not pass if the expression is a signal that is not type bool

rutefig:
> The test should pass if the expression is a query on a signal that is of type bool. I am not sure if this works with is_arith is_bool

What do you mean? Is this not supposed to check if the if condition is a boolean expression and throw an error if is not?

leolara:
is_arith and is_bool does not have access to the symbol table so it does not know the type of a signal.

so

`if a` where `a` is a signal is only valid if and only if `a` is of type `bool`. For the rest of cases it is ok to check is_bool, but in the future we should add another rule that checks the types and operators are compatible.

rutefig:
> is_arith and is_bool does not have access to the symbol table so it does not know the type of a signal.
> 
> so
> 
> `if a` where `a` is a signal is only valid if and only if `a` is of type `bool`. For the rest of cases it is ok to check is_bool, but in the future we should add another rule that checks the types and operators are compatible.

Cool I didn't think about it, but it makes sense, fixed right now. Now it is allowing only logical expressions and bool signals. Thank you for the observation 😄

src/compiler/semantic/rules.rs
```diff
@@ -336,11 +336,87 @@ fn types_rule_tl(
...
+                            "Signal in condition for the if statement must be bool, found {:#?}",
+                            expr
+                        ),
+                        dsym,
+                    )
+                }
+            } else {
+                analyser.error(
+                    format!(
+                        "Condition in if statement uses undeclared variable {:#?}",
```
leolara, line 377: this error is capture already in another checked, we should just not do anything if it is not find, as it is outside this scope of responsability

src/compiler/semantic/rules.rs
```diff
@@ -1062,4 +1138,77 @@ mod test {
...
+             }
+            }
+
+            // There is always a state called final.
+            // Output signals get automatically bound to the signals
+            // with the same name in the final step (last instance).
+            // This state can be implicit if there are no constraints in it.
+           }
+        ";
+
```
leolara, line 1201: You should also check that it passes with things like `3+i == a`; because that is a valid logical expression, I would just look at the top operator to see if it is logical or go with something more advance that recursively checks. But it is more complicated than currently. `==` root is a logical expression, but sides should be airthmetic expressions

rutefig: Oki I got it now! I was thinking that we could not use any field signal inside of the condition, that's why I was checking recursively. But what I understood now is that the rule just needs to make sure the condition should be a logic expr, true, false, or bool signal. Changed now for just checking the root expression of the condition

leolara (approved):

PR #257, opened 2024-06-03
Parameterize AssignmentGenerator by TraceGenerator

alxkzmn:


src/wit_gen.rs
```diff
@@ -114,12 +114,44 @@ impl<F> TraceContext<F> {
...
+/// ### Associated Types
+/// - `TraceGenerator::TraceArgs` The type of arguments that the trace generator takes.
+pub trait TraceGenerator<F> {
+    /// The type of arguments that the trace generator takes.
+    type TraceArgs: Clone;
+    /// Creates a new trace generator with the given trace function and number of steps.
+    /// ### Parameters
+    /// - `trace`: The trace function.
+    /// - `num_steps`: The number of steps in the circuit.
+    fn new(trace: Rc<Trace<F, Self::TraceArgs>>, num_steps: usize) -> Self;
```
leolara, line 147: So we are trying that different implementations could have a different way of generating TraceWitness, and if that case their constructor would be different

src/plonkish/ir/assignments.rs
```diff
@@ -125,18 +137,39 @@ impl<F: Field + Hash, TraceArgs> AssignmentGenerator<F, TraceArgs> {
         }
     }
 
+    /// Create an AssignmentGenerator without a trace generator
+    /// for the case when the trace is provided externally
+    pub fn without_trace_generator(
```
leolara, line 142: I don't think this method should be necessary, how would the AssignmentGenerator work without a TraceGenerator to provide the the TraceWitness?

src/wit_gen.rs
```diff
@@ -137,20 +169,35 @@ impl<F, TraceArgs> Default for TraceGenerator<F, TraceArgs> {
...
-
         (self.trace)(&mut ctx, args);
-
         ctx.get_witness()
     }
 }
 
+/// A no-op trace generator. The trace should be provided externally.
+#[derive(Clone, Default)]
+pub struct NoTraceGenerator;
```
leolara, line 187: This type should not be necessary

src/frontend/python/mod.rs
```diff
@@ -91,7 +91,7 @@ pub fn chiquito_ast_to_pil(witness_json: &str, rust_id: UUID, circuit_name: &str
 }
 
 fn add_assignment_generator_to_rust_id(
-    assignment_generator: AssignmentGenerator<Fr, ()>,
+    assignment_generator: AssignmentGenerator<Fr, NoTraceGenerator>,
```
leolara, line 94: So here we should pass a Python front-end specific TraceGenerator type that gets the data the way it comes from the python front-end

leolara:
Looking at the code, I think what we need is another TraceGenerator type, that takes as TraceArgs always TraceWitness, and then it can just return it in its own generate function.

Also, I think it would be better if MappingContext::map accepts the TraceGenerator type, but SuperCircuit as a whole does not depend on that, if possible

src/plonkish/compiler/mod.rs
```diff
@@ -88,7 +90,7 @@ pub fn compile_phase1<
             unit.columns.clone(),
             unit.placement.clone(),
             unit.selector.clone(),
-            TraceGenerator::new(Rc::clone(v), ast.num_steps),
+            TG::new(Rc::clone(v), ast.num_steps),
```
alxkzmn, line 93: Do I get it right that here the type of TG depends on the caller of `compile_phase1`? This is because `compile_phase1` is called from both python frontend and DSL. DSL is calling it via [`sub_circuit`](https://github.com/privacy-scaling-explorations/chiquito/blob/81cdc04991c776b498d0f1ee3704bff2490793ea/src/frontend/dsl/sc.rs#L45), python frontend is calling it via [`sub_circuit_with_ast`](https://github.com/privacy-scaling-explorations/chiquito/blob/81cdc04991c776b498d0f1ee3704bff2490793ea/src/frontend/dsl/sc.rs#L80). In [`chiquito_super_circuit_halo2_mock_prover`](https://github.com/privacy-scaling-explorations/chiquito/blob/81cdc04991c776b498d0f1ee3704bff2490793ea/src/frontend/python/mod.rs#L108) pythoin frontend is deserializing witness and calling [`map_with_witness`](https://github.com/privacy-scaling-explorations/chiquito/blob/81cdc04991c776b498d0f1ee3704bff2490793ea/src/plonkish/ir/sc.rs#L112). As I understand, you want to get rid of that too because the `TraceGenerator` should know where the witness is coming from. However, the witness was obtained [here from the `super_witness`](https://github.com/privacy-scaling-explorations/chiquito/blob/81cdc04991c776b498d0f1ee3704bff2490793ea/src/frontend/python/mod.rs#L132-L136) after both AssignmentGenerator and TraceGenerator were created. Does this mean that I need to refactor their creation and somehow pass the supercircuit witness JSON into them? E.g., by passing `super_witness` into `sub_circuit_with_ast` and `compile_phase1`?

alxkzmn:
@leolara I changed the naming a bit, SimpleTraceGenerator now has the TraceWitness as the arg type and returns it in generate(), and WitnessTraceGenerator does the actual generation by calling trace function.

src/wit_gen.rs
leolara: Please, check this @alxkzmn I don't think new should be in the trait

src/wit_gen.rs
```diff
@@ -137,20 +169,33 @@ impl<F, TraceArgs> Default for TraceGenerator<F, TraceArgs> {
...
         ctx.get_witness()
     }
 }
 
+#[derive(Clone, Default)]
+pub struct SimpleTraceGenerator;
+
+impl<F: Clone> TraceGenerator<F> for SimpleTraceGenerator {
+    type TraceArgs = TraceWitness<F>;
+    fn new(_trace: Rc<Trace<F, Self::TraceArgs>>, _num_steps: usize) -> Self {
```
leolara, line 190: This does not need a new method, Default is enought

src/plonkish/compiler/mod.rs
alxkzmn: @leolara I understand that naturally "new" feels like a constructor and shouldn't be in the trait, but this problem of who is instantiating the TraceGenerator is still unresolved. It is generic here so I could only put the "new" in the trait.

src/wit_gen.rs
alxkzmn: https://github.com/privacy-scaling-explorations/chiquito/pull/257#discussion_r1625666432

examples/keccak.rs
```diff
@@ -2210,25 +2211,28 @@ struct CircuitParams {
...
-    super_circuit::<F, KeccakCircuit, _>("keccak", |ctx| {
+    super_circuit::<F, KeccakCircuit, _, WitnessTraceGenerator<F>>("keccak", |ctx| {
         let in_n = (input_len * 8 + 1 + RATE_IN_BITS as usize) / RATE_IN_BITS as usize;
         let step_num = in_n * (1 + NUM_ROUNDS as usize);
 
         let single_config = config(SingleRowCellManager {}, SimpleStepSelectorBuilder {});
         // config(SingleRowCellManager {}, LogNSelectorBuilder {});
 
-        let (_, constants_table) = ctx.sub_circuit(
+        let (_, constants_table): (AssignmentGenerator<F>, LookupTable) = ctx.sub_circuit(
```
leolara, line 2221: why do we need to make the type explicit now?

alxkzmn: Because sub_circuit is generic by AssigmentGenerator, so we either need to put generics in sub_circuit or in the return type (return is shorter so I chose that).

src/plonkish/compiler/mod.rs
alxkzmn: @leolara I think I see one possible solution for this. What if we have a TraceGeneratorFactory "abstract factory" trait? Then we can pass the concrete implementations from the outside of the `sub_circuit -> compile_phase1` chain. These concrete implementations will know how to instantiate the corresponding `TraceGenerator` with or without arguments.

leolara: Let me send a commit, What we just need is to have in SBPIR the TraceGenerator as the field, instead of the closure

leolara (approved):

PR #264, opened 2024-07-02
Rutefig/231 applying cse to compiler

rutefig:


src/poly/cse.rs
```diff
@@ -1,134 +1,121 @@
...
+    F: Field + Hash,
+    V: Clone + Eq + PartialEq + Hash + Debug,
+    SF: SignalFactory<V>,
+>(
+    decomp: &mut ConstrDecomp<F, V>,
+    constr: Expr<F, V, ()>,
+    common_ses: &[Expr<F, V, HashResult>],
+    assignments: &VarAssignments<F, V>,
+    signal_factory: &mut SF,
+) -> Expr<F, V, ()> {
```
leolara, line 34: @rutefig why constr doesn't have HashResult? We should have done it at this point for all of them

leolara: This bottom up, what component you need?

So I would do a function that subsitute one common_se into a constr, instead of an array, make the loop outside.

i would have something like: (decomp: &mut ConstrDecomp<F, V>, constr: Expr<F, V, HashResult>, common_se: expr<F, V, HashResult>, assignments: &VarAssignments<F, V>, signal_factory: &mut SF)

leolara: Also, a better way to keep track of the signals that we have created to represent a common subexpression. Perhaps adding method to ConstrDecomp?

src/poly/cse.rs
```diff
@@ -1,134 +1,121 @@
...
-
-        let assignments: VarAssignments<Fr, Queriable<Fr>> = generate_random_assignment(&vars);
-
-        println!("Assignments: {:#?}", assignments);
-
-        for key in &keys {
-            assert!(assignments.contains_key(key));
+    } else {
+        // Otherwise, replace the subexpressions recursively
+        match constr {
```
leolara, line 57: @rutefig perhaps we can abstract this out, with a method that recieves a lambda in Expr. Like expr.apply_subexpressions(|ses| ... )

src/poly/cse.rs
rutefig: So I think the loop is better to be outside because this way I will check from higher degree to lower naturally and replacing as soon as it finds it. And I followed the same pattern as the other optimisations, this way the code is more easy to understand I think...

src/poly/mod.rs
```diff
@@ -14,38 +14,88 @@ pub mod mielim;
...
-pub trait ToExpr<F, V, M> {
+pub trait ToExpr<F, V, M: Meta> {
     fn expr(&self) -> Expr<F, V, M>;
 }
 
 pub trait ToField<F> {
     fn field(&self) -> F;
 }
 
+pub trait Meta: Clone + Default + Debug + Hash {
```
alxkzmn, line 25: Cool trick with attaching meta! I have a question: do you think there could be any other use for Meta in the future besides hashing? If not, we could rename it into "HashingStrategy", and rename "from_expr" into something like "compute_hash".

leolara: Yes, we will use meta to keep DebugSymRefs all the way to the SBPIR, so the MockProver on SBPIR can show good error messages

src/poly/cse.rs
leolara: I don't know which loop you mean. I think the function should substitute one sub expression at a time, to control which subsexpressions get substituted first from outside

src/poly/cse.rs
```diff
@@ -1,134 +1,37 @@
...
-        assignments.insert(queriable.clone(), value);
-    }
-
-    assignments
+use super::{ConstrDecomp, Expr, HashResult, SignalFactory};
+use std::{fmt::Debug, hash::Hash};
+
+pub fn replace_expr<F: Field + Hash, V: Clone + Eq + Hash + Debug, SF: SignalFactory<V>>(
+    expr: &Expr<F, V, HashResult>,
+    common_exprs: &[Expr<F, V, HashResult>],
```
leolara, line 8: @rutefig please put common_expr only, so we can first substitue the higher degree subexpressions. We also after substituting have to see if there are common sub expression that are not there anymore because there were part of the substituted subexpression

src/poly/mod.rs
alxkzmn: You mean something like
```rust
impl Meta for DebugSymRef {
    fn from_expr<F: Field + Hash, V: Clone + Eq + Hash>(
        expr: &Expr<F, V, ()>,
        assignments: &VarAssignments<F, V>,
    ) -> Self {
        //create a debug sym from expr and assignments
       ...
    }
}
```

alxkzmn: I'm not sure if I fully understand the purpose but my concern is that it looks like "anything factory" because `from_expr` only tells that "something could be made of an expression and assignments". Also, we have to be sure that we won't need the hash and the debug symbols at the same time as there is only one Meta in an expression.

leolara: I am not sure, I was going to dive into the details once I am sure about the algorithm itself. What I mean is the M type parameter in expression. We will use it for DebugSymRef and other things, like in this case for the CSE.

You can do mappings that change the content of M.

alxkzmn: This is what concerns me. It looks overly generic as it can be a different thing at different times.

src/poly/mod.rs
```diff
@@ -61,27 +111,76 @@ impl<F, V, M> Expr<F, V, M> {
...
+            _ => self.clone(),
         }
+    }
+}
 
-        let hash_result = HashResult {
-            hash: hasher.finish(),
-            degree: self.degree(),
-        };
+impl<F: Field, V: Clone, M: Meta> Expr<F, V, M> {
```
leolara, line 166: @rutefig why does M has to be of Meta for this method to work? We should not add traits to the types if they are not necessary

src/poly/mod.rs
```diff
@@ -61,27 +111,76 @@ impl<F, V, M> Expr<F, V, M> {
             Expr::MI(_, m) => m,
         }
     }
-}
 
-#[derive(Debug, Clone)]
-pub struct HashResult {
-    pub hash: u64,
-    pub degree: usize,
+    pub fn map_meta<N: Meta, Func: Fn(&M) -> N>(&self, f: Func) -> Expr<F, V, N> {
```
leolara, line 115: @rutefig why do you want the map to work only if M is of type Meta, why limit it to this?

src/poly/cse.rs
rutefig: And that's how is being done' I'm just passing a reference to the list of the already found common ses and then I'm checking for each expression it it is a common one or if it has some common subexpression. I just compute the common ses all before replacing it because it gives more control on which should be replaced or not

src/poly/mod.rs
rutefig: What is the issue of being too generic? If you just plug in some metadata that you need at a time and then when you don't need anymore you just remove it, I think it's ok to be generic. But I can be wrong

rutefig: > I'm not sure if I fully understand the purpose but my concern is that it looks like "anything factory" because `from_expr` only tells that "something could be made of an expression and assignments". Also, we have to be sure that we won't need the hash and the debug symbols at the same time as there is only one Meta in an expression.

but I do agree the name is not the best, I was at first calling `apply_meta` and also it should not depend on the `assignments` for everything, so I need to change that

leolara: Why would a `M` would need always assignemtns to be created?

I am not sure this is a method that we would need for any `M` on a expression

src/poly/cse.rs
rutefig: so what is happening actually is on the `compiler/cse` mod I'm identifying all of the common subexpressions for that step based on the minimum degree and minimum of occurrences (which we can define) and this method is doing what you're saying, so, we are replacing the `expr`, but I'm passing a reference to the identified common subexpressions just to check for that  `expr` if it's a common se or not, and if it's not I start to check for each subexpression recursively on that expr. Does it make sense? Because I feel what you're saying is what is here but maybe it's not the clearest way. So I can try to improve that if you want, but what is happening is exactly what you mentioned

src/poly/mod.rs
rutefig: > Why would a `M` would need always assignemtns to be created?
> 
> I am not sure this is a method that we would need for any `M` on a expression

Yeah it's not, I'm gonna change it

src/poly/cse.rs
leolara: Yes, I read the code. We should not pass all the common sub expresions, only one, and from outside decide which one we start substituting. And then we have to check the situation again.

leolara:
A more global issue here, we need that what is considered a common sub expression that can be is the best should be configurable.

We need a trace that have a method, that received the relevant info about the common sub-expression and returns a u32. `0` means that it should be ignored and other number is the score and the biggest score should be picked to subsitute.

The first implementation should be:
 * if degree is less than a configurable number return 0
 * if number of repetations is less than a configurable number return 0
 * return number of occurances * degree

leolara (changes requested):

rutefig:
> A more global issue here, we need that what is considered a common sub expression that can be is the best should be configurable.
> 
> We need a trace that have a method, that received the relevant info about the common sub-expression and returns a u32. `0` means that it should be ignored and other number is the score and the biggest score should be picked to subsitute.
> 
> The first implementation should be:
> 
> * if degree is less than a configurable number return 0
> * if number of repetations is less than a configurable number return 0
> * return number of occurances * degree



> A more global issue here, we need that what is considered a common sub expression that can be is the best should be configurable.
> 
> We need a trace that have a method, that received the relevant info about the common sub-expression and returns a u32. `0` means that it should be ignored and other number is the score and the biggest score should be picked to subsitute.
> 
> The first implementation should be:
> 
> * if degree is less than a configurable number return 0
> * if number of repetations is less than a configurable number return 0
> * return number of occurances * degree

Ok, I think that is easy, is just a matter of adding a new method for the `SubexprInfo` (which already keeps track of the degree and the number of occurrences) that will return the score. I can plugin again a CSEConfig, I only removed because I wanted to be sure of the core logic first for the replacements, but since it was done before I can add it again.

src/poly/mod.rs
```diff
@@ -61,27 +89,85 @@ impl<F, V, M> Expr<F, V, M> {
             Expr::MI(_, m) => m,
         }
     }
-}
 
-#[derive(Debug, Clone)]
-pub struct HashResult {
-    pub hash: u64,
-    pub degree: usize,
+    pub fn map_meta<N, Func: Fn(&M) -> N>(&self, f: Func) -> Expr<F, V, N> {
```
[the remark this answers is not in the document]
rutefig: Yes it is true, we can remove this function

src/compiler/cse.rs
```diff
@@ -0,0 +1,369 @@
...
+
+        let mut hashed_exprs = Vec::new();
+        for expr in exprs.iter() {
+            let hashed_expr = expr.hash(&rand_assignments);
+            hashed_exprs.push(hashed_expr);
+        }
+
+        let best_expr = find_optimal_subexpression(&hashed_exprs, &HashSet::new());
+
+        assert_eq!(format!("{:?}", best_expr.unwrap()), format!("{:?}", expr4));
```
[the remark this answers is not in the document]
rutefig: I did the tests mostly to check if it was working as expected and fix some issues, but I agree now is more interesting to do it a bit more complex to make sure everything is good.

src/poly/mod.rs
```diff
@@ -332,15 +418,42 @@ pub trait SignalFactory<V> {
...
+                Some((s, e))
+            } else {
+                None
+            }
+        })
+    }
+}
+
+impl<F: Clone, V: Clone> ConstrDecomp<F, V, HashResult> {
+    pub fn find_auto_signal_by_hash(&self, hash: u64) -> Option<(&V, &Expr<F, V, HashResult>)> {
```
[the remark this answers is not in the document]
rutefig: It is not being used right now, but I think it can be useful anytime, and this was to look into the `ConstrDecomp` and find if there was already an `auto_signal` created for that expression, because this was the most certain way to know if there was already a signal for that expression or not

src/poly/mod.rs
```diff
@@ -332,15 +418,42 @@ pub trait SignalFactory<V> {
...
-    pub constrs: Vec<Expr<F, V, ()>>,
+    pub constrs: Vec<Expr<F, V, M>>,
     /// Expressions for how to create the witness for the generated signals the orginal expression
     /// has be decomposed into.
-    pub auto_signals: HashMap<V, Expr<F, V, ()>>,
+    pub auto_signals: HashMap<V, Expr<F, V, M>>,
 }
 
-impl<F, V> Default for ConstrDecomp<F, V> {
+impl<F, V: Debug, M> ConstrDecomp<F, V, M> {
```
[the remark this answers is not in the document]
rutefig: This way we can use the `ConstrDecomp` with some M applied to the Expr. Doesn't make sense?

src/sbpir/mod.rs
```diff
@@ -331,31 +386,88 @@ impl<F> StepType<F> {
...
         self.transition_constraints.push(condition)
     }
 }
 
-impl<F: Clone + Eq + Hash> StepType<F> {
-    pub fn decomp_constraints<M>(&mut self, mut decomposer: M)
+impl<F: Field + Hash, M: Clone> StepType<F, M> {
+    pub fn with_meta<N: Clone, ApplyMetaFn>(&self, apply_meta: ApplyMetaFn) -> StepType<F, N>
+    where
+        ApplyMetaFn: Fn(&Expr<F, Queriable<F>, ()>) -> N + Clone,
```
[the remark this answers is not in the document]
rutefig: Yes I agree, at some point I did it but I don't know why I moved back again, and I think I thought it would be safer forcing to remove the meta first but I think is not so relevant actually. Thank you for pointing that

src/poly/cse.rs
```diff
@@ -1,134 +1,102 @@
...
-    }
+/// This function replaces a common subexpression in an expression with a new signal.
+pub fn replace_expr<F: Field + Hash, V: Clone + Eq + Hash + Debug>(
+    expr: &Expr<F, V, HashResult>,
+    common_se: &Expr<F, V, HashResult>,
+) -> (Expr<F, V, HashResult>, ConstrDecomp<F, V, HashResult>) {
+    let new_expr = replace_subexpr(expr, common_se);
 
-    result
+    (new_expr, ConstrDecomp::default())
```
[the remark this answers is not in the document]
rutefig: That is true, I got confused because I was using it before and then it was not necessary. Sorry about it

src/compiler/cse.rs
```diff
@@ -0,0 +1,369 @@
...
+            .unwrap()
+            .auto_signals
+            .values();
+
+        assert!(common_ses_found_and_replaced
+            .clone()
+            .any(|expr| format!("{:?}", &expr) == "(a * b)"));
+        assert!(common_ses_found_and_replaced
+            .clone()
+            .any(|expr| format!("{:?}", &expr) == "(e * f * d)"));
```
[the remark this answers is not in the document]
rutefig: I was missing the constraint, fixed

src/compiler/cse.rs
```diff
@@ -0,0 +1,369 @@
...
+                exprs.push(constraint.expr.clone());
+            }
+
+            // Find the optimal subexpression to replace
+            if let Some(common_expr) = find_optimal_subexpression(&exprs, &replaced_hashes) {
+                // Add the hash of the replaced expression to the set
+                replaced_hashes.insert(common_expr.meta().hash);
+                // Create a new signal for the common subexpression
+                let (common_se, decomp) =
+                    create_common_ses_signal(&common_expr, &mut signal_factory);
```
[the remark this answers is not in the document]
rutefig: I was missing the constr actually, fixed right now, so I would say we still need to use the ConstrDecomp right?

leolara:
Why is not passing the tests?

leolara:
> We need a trace that have a method, that received the relevant info about the common sub-expression and returns a u32. 0 means that it should be ignored and other number is the score and the biggest score should be picked to subsitute.

@rutefig sorry, there where it says `trace` it should have said `trait`. Please, let me know if you don't understand something.

An object that implements a trait should be passed to return the score.

When that is done, let me know, to review again

rutefig:
> Why is not passing the tests?

Just formatting, I will fix it when it would be ready for review

rutefig:
> > We need a trace that have a method, that received the relevant info about the common sub-expression and returns a u32. 0 means that it should be ignored and other number is the score and the biggest score should be picked to subsitute.
> 
> @rutefig sorry, there where it says `trace` it should have said `trait`. Please, let me know if you don't understand something.
> 
> An object that implements a trait should be passed to return the score.
> 
> When that is done, let me know, to review again

@leolara Can you review again pls?

src/compiler/cse.rs
```diff
@@ -0,0 +1,466 @@
...
+    let mut count_map = HashMap::<u64, SubexprInfo>::new();
+    let mut hash_to_expr = HashMap::<u64, Expr<F, Queriable<F>, HashResult>>::new();
+
+    // Extract all subexpressions and count them
+    for expr in exprs.iter() {
+        count_subexpressions(expr, &mut count_map, &mut hash_to_expr, replaced_hashes);
+    }
+
+    // Find the best common subexpression to replace
+    let common_ses = count_map
```
leolara, line 189: @rutefig do we need this? the scorer should return 0 for the ones that do not pass the minimum threshold. Also config.min_occurrence, or config.min_degre are configs of the scorer, not of the CSE

leolara:
@rutefig we need a default implementation of the scorer following this rules as explained before:

> The first implementation should be:

> if degree is less than a configurable number return 0
> if number of repetations is less than a configurable number return 0
> return number of occurances * degree

src/compiler/cse.rs
leolara: This was explained here:

> The first implementation should be:

> if degree is less than a configurable number return 0
> if number of repetations is less than a configurable number return 0
> return number of occurances * degree

Like that is part of the scorer, to check min repetitions and min ocurrences.

rutefig: I put it as the Cse config because at first I was thinking min degree is a crucial part for the cse, but I understand your point on giving that responsibility to the Scorer and I agree. Changed that, take a look whenever you can pls :)

leolara:
The scorer is good, now I need to review the rest. I have not checked the tests, but make sure they are comprehensive and check corner cases.

src/compiler/cse.rs
```diff
@@ -0,0 +1,452 @@
...
+        Expr, HashResult, VarAssignments,
+    },
+    sbpir::{
+        query::Queriable, sbpir_machine::SBPIRMachine, ForwardSignal, InternalSignal, StepType,
+    },
+    wit_gen::NullTraceGenerator,
+};
+
+#[derive(Clone, Debug)]
+pub struct CseConfig {
```
leolara, line 23: @rutefig we are commenting all public interfaces now, but perhaps this can be just pub(crate), many things here perhaps are just pub(crate) because they are only called internelly from the compiler.

src/compiler/cse.rs
```diff
@@ -0,0 +1,452 @@
...
+        // Turn all Expr<F, V, ()> into Expr<F, V, HashResult>
+        let mut step_type_with_hash = step_type.transform_meta(|expr| {
+            let hashed_expr = expr.hash(&random_assignments);
+            hashed_expr.meta().clone()
+        });
+
+        // Extract all the expressions from the step type
+        let mut exprs = Vec::<Expr<F, Queriable<F>, HashResult>>::new();
+
+        for constraint in &step_type_with_hash.constraints {
```
leolara, line 127: @rutefig perhaps this is better expressed as a map?

src/compiler/cse.rs
```diff
@@ -0,0 +1,452 @@
...
+    circuit
+}
+
+fn cse_for_step<F: Field + Hash, S: Scoring<F>>(
+    step_type: &mut StepType<F, ()>,
+    forward_signals: &[ForwardSignal],
+    config: &CseConfig,
+    scorer: &S,
+) {
+    let mut signal_factory = SignalFactory::default();
```
leolara, line 96: @rutefig in general this function is too long. You should see how to split it in many small functions, and make this function read more like a script of all the different steps that it take

src/compiler/cse.rs
```diff
@@ -0,0 +1,452 @@
...
+                .constraints
+                .iter_mut()
+                .for_each(|constraint| {
+                    constraint.expr = replace_expr(&constraint.expr, &common_se);
+                });
+        } else {
+            // No more common subexpressions found, exit the loop
+            break;
+        }
+        *step_type = step_type_with_hash.transform_meta(|_| ());
```
leolara, line 161: What is this? either comment or new method that is more clear what it does

src/compiler/cse.rs
```diff
@@ -0,0 +1,452 @@
...
+        let mut exprs = Vec::<Expr<F, Queriable<F>, HashResult>>::new();
+
+        for constraint in &step_type_with_hash.constraints {
+            exprs.push(constraint.expr.clone());
+        }
+
+        // Find the optimal subexpression to replace
+        if let Some(common_expr) = find_optimal_subexpression(&exprs, &replaced_hashes, scorer) {
+            // Add the hash of the replaced expression to the set
+            replaced_hashes.insert(common_expr.meta().hash);
```
leolara, line 134: Why do we need replaced_hashes? once it has been replaced, it will show only once in the constraints, and hence it will not be chosen. Right?

src/compiler/cse.rs
```diff
@@ -0,0 +1,452 @@
...
+            // Create a new signal for the common subexpression
+            let (common_se, decomp) = create_common_ses_signal(&common_expr, &mut signal_factory);
+
+            // Add the new signal to the step type and a constraint for it
+            decomp.auto_signals.iter().for_each(|(q, expr)| {
+                if let Queriable::Internal(signal) = q {
+                    step_type_with_hash.add_internal(*signal);
+                }
+                step_type_with_hash.auto_signals.insert(*q, expr.clone());
+                step_type_with_hash.add_constr(format!("{:?}", q), expr.clone());
```
leolara, line 144: Either expr is already q - common_ses and then autosignal should be q, common_ses or expr is common_ses and constraint should be 1-expr.

How is this not coming in the tests?

src/compiler/cse.rs
```diff
@@ -0,0 +1,452 @@
...
+    expr: &Expr<F, Queriable<F>, HashResult>,
+    count_map: &mut HashMap<u64, SubexprInfo>,
+    hash_to_expr: &mut HashMap<u64, Expr<F, Queriable<F>, HashResult>>,
+    replaced_hashes: &HashSet<u64>,
+) {
+    let degree = expr.degree();
+    let hash_result = expr.meta().hash;
+
+    // Only count and store if not already replaced
+    if !replaced_hashes.contains(&hash_result) {
```
leolara, line 223: @rutefig if it is already replaced, it should always count to one, so there is no need to special case

src/compiler/cse.rs
```diff
@@ -0,0 +1,452 @@
...
+    let mut count_map = HashMap::<u64, SubexprInfo>::new();
+    let mut hash_to_expr = HashMap::<u64, Expr<F, Queriable<F>, HashResult>>::new();
+
+    // Extract all subexpressions and count them
+    for expr in exprs.iter() {
+        count_subexpressions(expr, &mut count_map, &mut hash_to_expr, replaced_hashes);
+    }
+
+    // Find the best common subexpression to replace based on the score
+    let best_subexpr = count_map
```
leolara, line 198: If you iterate over the expressions instead of the count_map, you don't need hash_to_expr, because the expressions already contain the hash.

src/poly/cse.rs
```diff
@@ -1,134 +1,92 @@
...
+        common_se.clone()
+    } else {
+        // Recursively apply the function to the subexpressions
+        expr.apply_subexpressions(|se| replace_expr(se, common_se))
     }
+}
 
-    assignments
+/// This function creates a new signal for a common subexpression.
+pub fn create_common_ses_signal<
```
leolara, line 28: why do we need to return a ConstrDecomp? if we jsut return the signal is ok, because anyway outside you get what you put in ConstrDecomp and do other stuff with it

src/poly/mod.rs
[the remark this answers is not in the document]
leolara: I mean what is the method get_auto_signal for?

src/sbpir/mod.rs
```diff
@@ -231,6 +231,38 @@ impl<F, TG: TraceGenerator<F>> SBPIRLegacy<F, TG> {
     }
 }
 
+impl<F: Field + Hash, TG: TraceGenerator<F> + Clone, M: Clone> SBPIRLegacy<F, TG, M> {
```
leolara, line 234: Please, make CSE work on the new compiler, instead of in the legacy compiler

leolara (approved):

src/poly/mod.rs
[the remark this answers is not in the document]
rutefig: To get an `auto_signal` from its annotation, in the end I didn't use it anymore but I think it can be useful. Do you want to remove it?

src/compiler/cse.rs
rutefig: Yes it is true, this was used when the replacement was being done for more than one expression at a time. Sorry

src/compiler/cse.rs
rutefig: Can you check now? I fixed one redundant constraint that was being added which was from this line: `step_type_with_hash.add_constr(format!("{:?}", q), expr.clone());`

And I added the explicit test to make sure we have the constraint that checks the equality of the auto_signal with the common subexpression, which in this case is common_ses - q, equivalent to what you said.

What is there right now is that when I create the new signal it is created on the `ConstrDecomp` because before this was being done for more than one common subexpr. I didn't remove the `ConstrDecomp` because I thought it would be needed later for the witness generation but I might be wrong. Anyways it seems a bit more modular this way and the auto_eq method is being reused for applying directly the constraint but if you think it causes confusion I can refactor in a way that the  `ConstrDecomp` is not used anymore. Lmk your thoughts pls

src/poly/cse.rs
rutefig: can you check this comment pls regarding this? https://github.com/privacy-scaling-explorations/chiquito/pull/264#discussion_r1721563179

src/compiler/cse.rs
rutefig: But this `count_map` main purpose is to count the expressions and I think the counting should be outside of the expression itself because it should only live for each cse cycle. What I did was to store the needed information from the expression alongside with the count to be more straightforward

src/compiler/cse.rs
leolara: It is not needed for witness generation, when you add it to auto_signals that makes the witness generation. `ConstrDecomp` is not stored anywhere so it is not needed for this. It could be use like before for more than one. Now it reduces the readability because it seems it is necessary for something when it is not. You are putting this info into the structure and then getting it out using map, which makes it less clear that there is only one, for example.

src/poly/cse.rs
```diff
@@ -1,134 +1,92 @@
...
-        let value = F::random(&mut rng);
-        assignments.insert(queriable.clone(), value);
+    if expr.meta().degree < common_se.meta().degree {
+        // If the current expression's degree is less than the common subexpression's degree,
+        // it can't contain the common subexpression, so we return it as is
+        expr.clone()
+    }
+    // If the expression is the same as the common subexpression return the signal
+    else if expr.meta().hash == common_expr_hash {
+        common_se.clone()
```
leolara, line 20: You are not returning the signal here, but the actual subexpression.

src/compiler/cse.rs
rutefig: Ok, so agreed and removed 👍🏼

PR #268, opened 2024-07-09
Infer the k value from the circuit witness height

alxkzmn:


leolara:
@alxkzmn could you explain why you are changing the pre-processing? k is dependant of number of rows from plonkish IR that is known before setup without any assignation done

alxkzmn:
K cannot be known in advance because of the way the blinding factors are accounted for in Halo2. This is why here https://github.com/privacy-scaling-explorations/chiquito/blob/8246537b59a25a6966087c93fee44ad4eae6ee9d/src/plonkish/backend/halo2.rs#L88 we need to add the minimum number of rows that are required by the constraint system. The reason is that the more advice queries there are, the more blinding factors a circuit should have.

alxkzmn:
See
https://github.com/zcash/halo2/blob/7df93fd855395dcdb301a857d4b33f37903bbf76/halo2_proofs/src/plonk/circuit.rs#L1464-L1472
https://github.com/zcash/halo2/blob/7df93fd855395dcdb301a857d4b33f37903bbf76/halo2_proofs/src/plonk/circuit.rs#L1435-L1460

leolara:
@alxkzmn ok, great understanding of halo2 internals!! I am going to see if I can merge this branch into mine without much problems

leolara:
I have merged this code into https://github.com/privacy-scaling-explorations/chiquito/pull/269

PR #276, opened 2024-07-18
Remove chiquito2halo2 function in halo2 backend

alxkzmn:


src/plonkish/compiler/mod.rs
```diff
@@ -54,11 +54,20 @@ pub fn compile<
     }
 }
 
-pub struct PlonkishCompilationResult<F, TG> {
+pub struct PlonkishCompilationResult<F: Clone, TG: TraceGenerator<F> + Clone> {
     pub circuit: Circuit<F>,
     pub assignment_generator: Option<AssignmentGenerator<F, TG>>,
 }
 
+impl<F: Clone, TG: TraceGenerator<F> + Clone> Clone for PlonkishCompilationResult<F, TG> {
```
leolara, line 62: @alxkzmn I think you can do an automatic derivation

src/plonkish/compiler/mod.rs
```diff
@@ -54,11 +54,20 @@ pub fn compile<
     }
 }
 
-pub struct PlonkishCompilationResult<F, TG> {
+pub struct PlonkishCompilationResult<F: Clone, TG: TraceGenerator<F> + Clone> {
```
leolara, line 57: @alxkzmn I don't think you need to make the generic types Clone here, just in the impl clause that needs that

leolara:
@alxkzmn

now the halo2 mock prover is not a mock prover, anymore?

The changes to python, apart random setup, are just linting?

alxkzmn:
Yep, it's not a mock prover anymore but I didn't want to rename it everywhere because we're still waiting for https://github.com/privacy-scaling-explorations/halo2/pull/356 to be merged so we can have the "mock prover" again. Those big changes to Poseidon example is because of formatting, the only real change there is passing the `params_path` value.

leolara (approved):

PR #277, opened 2024-07-21
Refactor lookup argument creation

alxkzmn:
The approach is to determine the fixed column type "lazily" before allocation by detecting whether it is a single fixed query or not. This allows for more flexibility (to keep the halo2 imported columns as they are; not to have to decide between table columns and fixed columns in advance).


src/plonkish/backend/halo2.rs
```diff
@@ -258,14 +264,51 @@ impl<F: PrimeField + From<u64> + Hash> ChiquitoHalo2<F> {
...
-                let mut exprs = Vec::new();
-                for (src, dest) in lookup.exprs.iter() {
-                    exprs.push((self.convert_poly(meta, src), self.convert_poly(meta, dest)))
+
+            let mut single_fixed_queries = vec![];
+            let mut other_queries = vec![];
+            for (idx, (_, dest)) in lookup.exprs.iter().enumerate() {
+                match dest {
+                    PolyExpr::Query((column, _, _), _) => {
+                        if column.ctype == cFixed
```
alxkzmn, line 273: @leolara here I added the explicit fixed column type check but this may be unnecessary if we know for sure that lookups only involve fixed columns.

src/plonkish/backend/halo2.rs
```diff
@@ -804,3 +884,38 @@ pub(crate) fn chiquito2Halo2<F: PrimeField + From<u64> + Hash>(
 ) -> ChiquitoHalo2<F> {
     ChiquitoHalo2::new(circuit)
 }
+
+impl<F> ExpressionWithColumn for Expression<F> {
+    fn involves(&self, column: &cColumn) -> bool {
```
leolara, line 889: @alxkzmn why do we need this method in both types of expressions?

leolara: Because of the imported expressions ?

src/plonkish/backend/halo2.rs
leolara: yes, but this seems more flexible

src/plonkish/backend/halo2.rs
alxkzmn: Yes, it is necessary because we have `PolyExpr::Halo2Expr` here: https://github.com/privacy-scaling-explorations/chiquito/blob/4b077b2b598c42d2cd5e2c5dc1ec5b79c30c96f5/src/plonkish/ir/mod.rs#L194

src/plonkish/backend/halo2.rs
```diff
@@ -258,14 +264,51 @@ impl<F: PrimeField + From<u64> + Hash> ChiquitoHalo2<F> {
...
-                for (src, dest) in lookup.exprs.iter() {
-                    exprs.push((self.convert_poly(meta, src), self.convert_poly(meta, dest)))
+
+            let mut single_fixed_queries = vec![];
+            let mut other_queries = vec![];
+            for (idx, (_, dest)) in lookup.exprs.iter().enumerate() {
+                match dest {
+                    PolyExpr::Query((column, _, _), _) => {
+                        if column.ctype == cFixed
+                            || column.ctype == Halo2Fixed && self.has_single_fixed_query(column)
```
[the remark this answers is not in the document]
alxkzmn: Yes, we can add the TableColumn import API for the imported Halo2 columns and this will eliminate the need for figuring out which is the column type.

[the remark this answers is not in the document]
leolara: What is the TableColumn import API?

I suggest to add a new method to the dsl that allows to import halo2 table columns, and a new type of query for them, like ortogonal to fixed.

src/plonkish/backend/halo2.rs
leolara: Ok

src/frontend/dsl/mod.rs
```diff
@@ -82,6 +82,12 @@ impl<F, TG: TraceGenerator<F>> CircuitContext<F, TG> {
         Queriable::Halo2FixedQuery(self.circuit.add_halo2_fixed(name, column), 0)
     }
 
+    /// Imports a halo2 table column with a name string into the circuit and returns a
+    /// `Queriable` instance representing the imported column.
+    pub fn import_halo2_table(&mut self, name: &str, column: TableColumn) -> Queriable<F> {
+        Queriable::Halo2FixedQuery(self.circuit.add_halo2_table(name, column), 0)
```
leolara, line 88: @alxkzmn there should be also a new type of query, Halo2TableQuery. Why not?

src/plonkish/backend/halo2.rs
```diff
@@ -120,7 +121,7 @@ pub struct ChiquitoHalo2<F: PrimeField + From<u64>> {
     pub(crate) plonkish_ir: Circuit<F>,
 
     advice_columns: HashMap<UUID, Column<Advice>>,
-    fixed_columns: HashMap<UUID, Column<Fixed>>,
+    fixed_columns: HashMap<UUID, (Column<Fixed>, Option<TableColumn>)>,
```
leolara, line 124: @alxkzmn perhaps can we do the distinction already before filling this struct, so instead of the option we would have two hashmaps?

leolara: Or this complicates the code?

leolara: or it seems that allocate_fixed_columns should return two hashmaps, one for fixed and another for table

alxkzmn: Yes, this is possible.

src/plonkish/backend/halo2.rs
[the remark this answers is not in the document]
alxkzmn: You mean this? https://github.com/privacy-scaling-explorations/chiquito/blob/b36689c3b4cc8450d55d611ba782c4e8e9e2b8dc/src/frontend/dsl/mod.rs#L87 It's already there

src/frontend/dsl/mod.rs
alxkzmn: Regarding this - I have been trying a different approach that is not using any Halo2 frontend code at all. Right now we need the TableColumns only because we are still using the CosntraintSystem from the Halo2 frontend. Later it is converted to CosntraintSystemMid which is in the middleware. It is possible to create the CosntraintSystemMid directly by mapping the chiquito poly expressions and lookups to halo2 equivalents (that is currently happening inside the frontend ConstraintSystem). I have consulted with Edu and he thinks that it's feasible without writing much new code (also there are some examples of how it's done from plonky3 frontend to halo2 middleware).

leolara: So, this makes me confused. I thought we were eliminating the halo2 front-end entirely. I am unsure what we should do. If we are going to still keep using ContraintSystem then I think the approach I am describing is the best, to separate the columns between fixed and table all the way.

leolara: So, in that case we only need to do the guessing for the native chiquito fixed columns

alxkzmn:
This may be relevant if we're planning to maintain the Halo2 frontend compilation path. But if we're fully deprecating it we don't need to worry about lookup sanity checks and TableColumns.

alxkzmn:
Yes.

PR #282, opened 2024-07-25
Remove the Halo2 frontend ConstraintSystem

alxkzmn:
This PR is a better approach to Halo2 middleware compilation than #277. It eliminates the need for TableColumns that are a purely frontend concept. If we choose to proceed with this path, we can completely remove the dependency on Halo2 frontend. I have prepared the corresponding changes here: https://github.com/privacy-scaling-explorations/chiquito/commit/93f955301111b377412cea8eaca3d3b8c2b4c585


leolara:
@alxkzmn let's get this one that the imported columns and expressions are still from halo2 frontend and gives an error if there are. The rest is cool that it does not depend of the halo2 frontend at all

src/plonkish/backend/halo2.rs
```diff
@@ -187,193 +187,207 @@ impl<F: PrimeField + From<u64> + Hash> ChiquitoHalo2<F> {
...
             fixed_columns: Default::default(),
-            instance_column: Default::default(),
             ir_id,
         }
     }
 
-    fn configure_columns_sub_circuit(&mut self, meta: &mut ConstraintSystem<F>) {
-        let mut advice_columns = HashMap::<UUID, Column<Advice>>::new();
-        let mut fixed_columns = HashMap::<UUID, Column<Fixed>>::new();
+    fn configure_halo2_columns(&mut self, meta: &mut ConstraintSystemBuilder<F>) {
```
leolara, line 194: @alxkzmn why is called meta here but cs_builder somewhere. meta was the name we were using in the times of zkevm

src/plonkish/backend/halo2.rs
```diff
@@ -187,193 +187,207 @@ impl<F: PrimeField + From<u64> + Hash> ChiquitoHalo2<F> {
...
             });
         }
+
+        self.advice_columns = advice_columns;
+        self.fixed_columns = fixed_columns;
     }
+}
 
-    fn convert_poly(&self, meta: &mut VirtualCells<'_, F>, src: &PolyExpr<F>) -> Expression<F> {
+impl<F: PrimeField> ConstraintSystemBuilder<F> {
```
leolara, line 244: @alxkzmn I think all the `impl` of ConstraintSystemBuilder should be after the `struct` defintion

leolara (approved):
A few style comments but approved

src/plonkish/backend/halo2.rs
alxkzmn: I'll change it to constraint system too, meta is not a very meaningful name but it is used in Halo2 code.

PR #284, opened 2024-08-06
Refactor compiler to return a new structure for multiple machines

alxkzmn:


alxkzmn:
@leolara should I implement the instantiation of the new SBPIR in this PR?

leolara:
@alxkzmn it is good idea to split tasks into smaller PRs as you are doing here :-)

leolara (approved):

PR #289, opened 2024-08-22
Expose debug symbol start and end for the language server

alxkzmn:
We have accidentally made the start and end of debug symbol private, so here's a fix for the language server. Also had to update the halo2 to the latest version because the language server didn't want to compile (there was a missing bound on a variable generic inside the halo2 middleware `Expression`, not sure why Chiquito itself compiled successfully).

### DIFF
--- a/src/parser/ast/mod.rs
+++ b/src/parser/ast/mod.rs
@@ -142,12 +142,14 @@ impl PartialEq for DebugSymRef {
 }
 
 /// Interface of the debug symbol reference to use with the Chiquito language server
-pub trait LanguageServerInterface {
+pub trait Spanned {
+    /// Get span start
     fn get_start(&self) -> usize;
+    /// Get span end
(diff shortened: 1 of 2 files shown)


rutefig (approved):
lgtm

src/parser/ast/mod.rs
```diff
@@ -141,6 +141,22 @@ impl PartialEq for DebugSymRef {
     }
 }
 
+/// Interface of the debug symbol reference to use with the Chiquito language server
+pub trait LanguageServerInterface {
```
leolara, line 145: @alxkzmn I don't think it is custom in Rust to call Interface to a trait. Also, the interface is used by the Language Server, not an interface to the Language Server.

leolara (changes requested):

leolara (approved):

PR #295, opened 2024-08-26
Parse HyperTransition syntax

alxkzmn:


src/parser/chiquito.lalrpop
```diff
@@ -37,9 +37,9 @@ pub Statements: Vec<Statement<BigInt, Identifier>> = {
 }
 
 ParseStatement: Statement<BigInt, Identifier> = {
-    <l: @L> <s: StatementType> ";" <r: @R> => add_dsym(dsym_factory.create(l,r), s),
```
alxkzmn, line 40: Each statement type already has its own L and R parsing, so `add_dsym` wasn't needed anymore.

src/parser/chiquito.lalrpop
```diff
@@ -111,6 +113,14 @@ ParseTransition: Statement<BigInt, Identifier> = {
     <l: @L> "->" <id: Identifier> <block: StatementsBlock> <r: @R> => build_transition(dsym_factory.create(l,r), id, block),
 }
 
+HyperTransition: Statement<BigInt, Identifier> = {
+    <l: @L> <call: Call> "->" <st: Identifier> <r: @R> => build_hyper_transition(dsym_factory.create(l,r), st, call),
+}
+
+Call: Statement<BigInt, Identifier> = {
+    <l: @L>  <ids: ParseIdsList> "<==" <machine: Identifier> "(" <es:ParseExpressionList> ")"  <r: @R> => Statement::Call(dsym_factory.create(l,r), ids, machine, es),
```
leolara, line 121: As a call can be part of an expression in general, probably in the future when we implement actual function calls, it should be an Expression and not an statement @alxkzmn

alxkzmn: Got it, will implement.

leolara (approved):
There are some todos, that we should put in issues if they are not already. Perhaps you can put that they need spec'ing :-)

Approved

